PR #16, opened 2022-12-16
Add a new API `Wallet::new_with_online()`.

monaka:
I'm developing RESTful http interface for RGB-Lib.
This PR is for my app.

As you might know, web frameworks tends to be stateless.
So I want to keep my Wallet instance over sessions.
But as far as I trided, the instance of Wallet structure couldn't store in reference count traits like `Rc<>` or `Arc<>`. 

I also tried instancing `Wallet` per session.
Results were "panicked" because of thread generation in `_go_online()`.

This PR enables to inject `Online` object when a Wallet instance is created.
This approach is kept a backward compatibility. All tests are passed.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -671,6 +671,11 @@ pub struct Wallet {
 impl Wallet {
     /// Create a new RGB wallet based on the provided [`WalletData`]
     pub fn new(wallet_data: WalletData) -> Result<Self, Error> {
+        Self::new_with_online(wallet_data, None)
+    }
+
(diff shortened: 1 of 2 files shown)


monaka:
After tracking #17, I realized another approach is required to fix my pain point.
I close this PR as "won't fix".

monaka:
@zoedberg Thank you for you suggestion.
I got the reason why this PR is useless before closing this.
Your comment makes me more clear.

And in fact, I tried to share an instance of `struct Wallet` before.
It was failed by another error message.
I'm going to create a tiny crate that reproduces errors.
I'll share it within tomorrow.

PR #20, opened 2022-12-18
Use the path `/tmp/tests/tmp` on Docs.rs.

### LINKED ISSUE #19: Build fails on docs.rs
image

I fixed a similar issue on another crate before.
I'm going to make a pull request soon.

monaka:
See also: https://docs.rs/about/builds#detecting-docsrs
          https://docs.rs/about/builds#cross-compiling

This will fix #19.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ license = "MIT"
 description = "RGB wallet library"
 exclude = ["migration"]
 
+[package.metadata.docs.rs]
+rustc-args = ["--cfg", "docsrs"]
+
 [workspace]
 members = [".", "migration"]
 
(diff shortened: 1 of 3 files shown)


monaka:
I did:
* self-review on my text editor. 
* check the shell script.

I've not:
* run on docs.rs CI.

Please do cross-check before merge.

zoedberg:
@monaka I'm not sure this will fix #19. Logs reported there are quite clear: it is not building because of rgb_node requiring write access to the file system (same as was happening with store daemon). I've already opened a PR for that: https://github.com/RGB-WG/rgb-node/pull/224

monaka:
@zoedberg I read a patch in RGB-WG/rgb-node#224.
And I agree that the root cause is write access permission.

This patch is not needed if a documentation process doesn't spawn `tests/start_services.sh` ... ah, it will be spawned only in `cfg(test)`.
I see. I close this.

PR #24, opened 2023-01-19
Update dependency.

monaka:
* Use `tempfile` instead of `tempdir` ( see also: https://rustsec.org/advisories/RUSTSEC-2018-0017.html ).
* Apply `cargo update` to reduce security risks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -227,7 +227,7 @@ dependencies = [
  "slab",
  "socket2",
  "waker-fn",
- "windows-sys 0.42.0",
+ "windows-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


monaka:
Note: I sent PR to rgb_node for fixing similar issues. https://github.com/RGB-WG/rgb-node/pull/240

zoedberg:
This PR breaks documentation so closing it, but thanks @monaka for letting us know that we need to change that (will do that in the next release that will happen very soon).

Generally speaking, please note that usually it's not possible to change a dependency without also updating code (unless dependency is unused, but in that case it could be removed).

PR #26, opened 2023-06-20
Remove and ignore `migration/Cargo.lock`.

monaka:
Because there creates a library.
More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html


zoedberg:
Commit says that you're removing and ignoring `migration/Cargo.lock`, but you've actually removed the `Cargo.lock` in the project root, not the one in `migration/`.

Moreover we're aware that libraries don't need to commit lock files, but we still find them useful for tests, ffi binaries and to keep track of last used dependencies among team members. They should not cause any problem, so I'll close this issue without merging it.

Issue #29, opened 2023-09-29
Store scrypt parameters alongside encrypted backup


St333p:
The current backup codebase uses default parameters provided for scrypt to derive the backup encryption key from password. However, those parameters might change in the future, as they adapt to hardware advancements; this could prevent decryption of old backups due to differing key generation parameters.

Scrypt parameters should be stored in cleartext along with the salt, as they are also required to reconstruct the key deterministically, and thus decrypt the backup.

zoedberg:
Solved in #30

PR #30, opened 2023-10-06
Store scrypt params along with backup 

St333p:
To be future-proof agains changes in recommended scrypt parameters for the backup, they should be stored as well. See #29 for details.


zoedberg (approved):
LGTM, thank you

PR #45, opened 2024-01-29
update strict_encoding

eauxxs:


Cargo.toml
```diff
@@ -50,7 +62,7 @@ rgb-schemata = "=0.10.1"
 rgb-std = "=0.10.9"
 rgb-wallet = "=0.10.9"
 rgb_core = { package = "rgb-core", version = "=0.10.8" }
-strict_encoding = "=2.6.1"
+strict_encoding = "2.6.2"
```
zoedberg, line 65:
```suggestion
strict_encoding = "=2.6.2"
```
We prefer to keep this dependency version locked

zoedberg (changes requested):
LGTM but please sign the commit and apply the small requested change. Thanks

zoedberg (changes requested):
Please squash commits into a single one. The commit "lock version" is not changing anything from master, it's fixing your previous commit, so it could be confusing

Issue #37, opened 2023-11-28
add bool for exact expiry on receive APIs


nicbus:
Let's add the `exact_expiry` bool parameter to receive APIs (`blind_receive` and `witness_receive`) that determines how an expired transfer should be handled.

When `true`, an expired transfer should be rejected.
When `false`, an expired transfer should be accepted anyway.

The current implementation behaves as if this new bool parameter was set to `false`.

This means that in the `_handle_expired_transfers` method (called by all APIs that require allocations), instead of immediately setting an expired transfer to the failed status, rgb-lib checks for one last time whether the sender eventually sent the assets and, if so, it accepts the assets even if the transfer has expired. The new behavior will change this when the bool parameter is set to true, by directly setting the expired transfer to the failed status, without checking one last time if the sender sent something.

A similar behavior is seen in the `fail_transfers` API where, before setting an expired transfer to the failed status, the library performs one last refresh. The updated library should also check whether this new bool parameter has been set to true and, in that case, fail the transfer without refreshing.

GylmarGonzalez:
Hi I am Gylmar Gonzalez, I am student in planb network node cubo plus El Salvador, I was assigned this issue , You can 
give me permision to colaborate in this project, because I clone de repo and I create my branch localy but I need create my branch remote.

Because I used diferente pcs , for my its more easy.

I have this error, when I try to push my new branch called issue37.

remote : permission to RGB-Tools/rgb-lib denied to GylmarGonzalez

Thanks.

zoedberg:
Hi @GylmarGonzalez! In order to contribute to the project you need to fork the project and then clone your fork. You will be then able to create and push branches. Then after pushing the branch you'll see on github the possibility to create a pull request with your branch (see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo and https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork).

GylmarGonzalez:
Hi.

I start to work in this issue I write this file its a my documentation, It will be updated as the solution progresses.

https://docs.google.com/document/d/1_Vvq1NLFCKzUderG8AzQuSmmRO3nhSxY_P2IKbqOEwc/edit?usp=drive_link

I am doing the migration of database with the new field in the table , I read the explanation in this link.

https://github.com/RGB-Tools/rgb-lib/pull/38

[...]

PR #47, opened 2024-04-09
retrying to implement exact expiry

arkanoider:
Hi @zoedberg ,

i am back, will do another tentative :)

I am restarting from scratch following the old ideas you exposed me in PR, this is not compiling and still here to ask you if i am doing right.

First of all i had a compiler error that on my machine seems created by the missing of this feature:

https://github.com/arkanoider/rgb-lib/blob/06bdd89cc5c079f049a8d195a4f80ca99e5d1cda/migration/Cargo.toml#L29

when i raised this db commands ( the second one if I remember correctly in particular) :

```docker pull postgres:latest

docker run -p 127.0.0.1:5432:5432/tcp --name migration-postgres \
    -e POSTGRES_PASSWORD=mysecretpassword -d postgres

DATABASE_URL=postgres://postgres:mysecretpassword@localhost:5432 \
    sea-orm-cli migrate refresh

DATABASE_URL=postgres://postgres:mysecretpassword@localhost:5432/postgres \
    sea-orm-cli generate entity -o src/database/entities --expanded-format

docker rm -f migration-postgres
```

After this I started updating the `test_witness_receive` with the false boolean for exact expiry, but there's a point where going down in functions call seems to me that i need to update this structure:

```Rust
#[derive(Clone, Eq, PartialEq, Debug)]
pub struct RgbInvoice {
    pub transports: Vec<RgbTransport>,
    pub contract: Option<ContractId>,
    pub iface: Option<TypeName>,
    pub operation: Option<TypeName>,
    pub assignment: Option<FieldName>,
    pub beneficiary: Beneficiary,
    pub owned_state: TypedState,
    pub chain: Option<Chain>,
    /// UTC unix timestamp
    pub expiry: Option<i64, bool>,
    pub unknown_query: IndexMap<String, String>,
}
```

called inside `_receive` method. This is placed in the crate [rgb-wallet](https://crates.io/crates/rgb-wallet) so maybe I should also touch that structure, but maybe I am wrong...

Can you confirm me that's the right path?

Still sorry for my noobnees, but I am trying to learn more about rgb, but my time is scarse, so I am more going around the code to learn, but I am missing a bit of the bigger scenario of rgb.

Thanks for your patience!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -158,6 +158,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d96bd03f33fe50a863e394ee9718a706f988b9079b20c3784fb726e7678b62fb"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8901269c6307e8d93993578286ac0edf7f195079ffff5ebdeea6a59ffb7e36bc"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c75ac65da39e5fe5ab759307499ddad880d724eed2f6ce5b5e8a26f4f387928c"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e28923312444cdd728e4738b3f9c9cac739500909bb3d3c94b43551b16517648"
+dependencies = [
+ "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "anstyle-wincon"
+version = "3.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cd54b81ec8d6180e24654d0b371ad22fc3dd083b6ff8ba325b72e00c87660a7"
+dependencies = [
+ "anstyle",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "arrayref"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


zoedberg:
Hi @arkanoider, we've just released 0.3.0-alpha.1 with support for RGB 0.11 beta 5 so I suggest rebasing this on top of that before proceeding (a lot of things changed).

> First of all i had a compiler error that on my machine seems created by the missing of this feature:

Thanks for reporting this, it has been fixed in recent commits

> Can you confirm me that's the right path?

No, the `RgbInvoice` structure should not be updated because the feature described in https://github.com/RGB-Tools/rgb-lib/issues/37 is just related to rgb-lib behavior on how to handle an expired invoice on the receiver side and doesn't require changes to its dependencies. In the `_receive` method we create an `RgbInvoice` just to make receive APIs return an invoice string that they can share to receive assets, but the sender doesn't need to know how the receiver decided to handle the expiration of this invoice, therefore this information doesn't need to be encoded in the `RgbInvoice` structure.

> Still sorry for my noobnees, but I am trying to learn more about rgb, but my time is scarse, so I am more going around the code to learn, but I am missing a bit of the bigger scenario of rgb.

No worries, I'm happy that you want to learn more about rgb and I'm happy to help. I hope my answer is clear enough, otherwise don't hesitate to ask more questions.

P.S. please next time just force push the branch of the existing PR instead of creating a new one :)

arkanoider:
> Hi @arkanoider, we've just released 0.3.0-alpha.1 with support for RGB 0.11 beta 5 so I suggest rebasing this on top of that before proceeding (a lot of things changed).

Sounds clear will do!


> Thanks for reporting this, it has been fixed in recent commits

You're welcome, glad that I was right. Thought it was something related to my configuration.

> No, the RgbInvoice structure should not be updated because the feature described in https://github.com/RGB-Tools/rgb-lib/issues/37 is just related to rgb-lib behavior on how to handle an expired invoice on the receiver side and doesn't require changes to its dependencies. In the _receive method we create an RgbInvoice just to make receive APIs return an invoice string that they can share to receive assets, but the sender doesn't need to know how the receiver decided to handle the expiration of this invoice, therefore this information doesn't need to be encoded in the RgbInvoice structure. 

Ok! So i can stay inside inside rgb-lib and not touch anything outside! Will proceed in that way in the next days in my spare time.

> P.S. please next time just force push the branch of the existing PR instead of creating a new one :)

You're fuckin right! I did some shit and restarted, but forcing a push was the thing to do!

arkanoider:
Hi @zoedberg ,

sorry for my silence...:) My fiatjob is killing me in these days, in the next days I will come back on Rgb again!

nicbus:
not relevant anymore (see #37), closing

PR #38, opened 2023-12-05
Testing exact expiry

arkanoider:
Hi,

pardon my noobness first of all.

Was trying to learn and contribute: this is what could be a starting point for issue #37 

Let me know how far I am from the right idea.

PR #54, opened 2024-12-13
Change the field of expiry 1.0

GylmarGonzalez:
I change the Transfer struct.

I run the database instruction.

I change the field from_db_transfer.


migration/src/lib.rs
```diff
@@ -1,12 +1,13 @@
...
+
+
+pub struct Migrator;
+
+#[async_trait::async_trait]
+impl MigratorTrait for Migrator {
+    fn migrations() -> Vec<Box<dyn MigrationTrait>> {
+        vec![Box::new(m20230608_071249_init_db::Migration)]
+    }
+}
```
zoedberg, line 13: please revert changes to this file

migration/src/m20230608_071249_init_db.rs
```diff
@@ -83,6 +83,7 @@ impl MigrationTrait for Migration {
                     .col(big_unsigned(BatchTransfer::CreatedAt))
                     .col(big_unsigned(BatchTransfer::UpdatedAt))
                     .col(big_unsigned_null(BatchTransfer::Expiration))
+                    .col(boolean(BatchTransfer::ExactExpiry))
```
zoedberg, line 86:
```suggestion
                    .col(boolean_null(BatchTransfer::ExactExpiry))
```
this should be nullable. 

please read carefully the comment I pointed out (https://github.com/RGB-Tools/rgb-lib/pull/38#issuecomment-1843346385), there I said
> as you see the new parameter is nullable

zoedberg (changes requested):
@GylmarGonzalez The reason you are encountering build errors (that you mentioned in https://github.com/RGB-Tools/rgb-lib/issues/37#issuecomment-2540560254) is that after running the migration you kept some changes that are unrelated to the change you did to the `migration/src/m20230608_071249_init_db.rs` file. Please carefully read the migration README, at the end it says:

> The command to generate entities will apply some unwanted changes, for example it will change the enum fields to integers and will remove some extra derives that we manually added. Those changes will need to be discarded, so please be sure to add only the code that is related to the new changes you just applied. To do this we suggest to first refresh the migration and generate entities with sea-orm-cli on the branch you are about to apply the DB changes on. The generated diff will only include unwanted changes, so they can be used as a reference to revert them.

GylmarGonzalez:
> @GylmarGonzalez The reason you are encountering build errors (that you mentioned in [#37 (comment)](https://github.com/RGB-Tools/rgb-lib/issues/37#issuecomment-2540560254)) is that after running the migration you kept some changes that are unrelated to the change you did to the `migration/src/m20230608_071249_init_db.rs` file. Please carefully read the migration README, at the end it says:
> 
> > The command to generate entities will apply some unwanted changes, for example it will change the enum fields to integers and will remove some extra derives that we manually added. Those changes will need to be discarded, so please be sure to add only the code that is related to the new changes you just applied. To do this we suggest to first refresh the migration and generate entities with sea-orm-cli on the branch you are about to apply the DB changes on. The generated diff will only include unwanted changes, so they can be used as a reference to revert them.


Hi its very confused.

maybe the better its create a new branch? I need to understand how to run the project without apply my changes.

1. I create a new branch.
2. I build the proyect? its not necesary run the first migration on database? or yes?
3. If I do this step previous , I create the database , this database not contain my changes, but If I build the proyect , will working? not showing errors?.

later the proyect it clean and working I doing my changes (I dont say the proyect its not working I only say I need know how to build the project..

Its confused when you say I dont need the do the migration, but I used the first file on database and I write my changes here, I dont know write the new field on database in creation or in other instruccion maybe do the alter of table in the same file initial of database.

I don't understand, how to revert the changes when you say "Those changes will need to be discarded,"
I think And only add new field and changed the struct.

Surely all these doubts are because I am learning rust.

zoedberg:
> maybe the better its create a new branch?

you don't need to create a new branch, if you want to start from scratch you can just hard reset the branch to the master branch. please study how git and github work because it's essential in order to do the task

> Its confused when you say I dont need the do the migration

I never said you don't need to do the migration. I said "You shouldn't generate a new migration file but edit the init one (since the library is still in alpha we do not need to support migrations from existing DBs)". This means you don't need to create a new migration file (with `sea-orm-cli migrate generate <migration_name>`) and you don't have to write an alter table statement. You need to change the `migration/src/m20230608_071249_init_db.rs` file, as you did in this PR, but as suggested on the code the field needs to be a nullable boolean.
Then, after editing the init file, you'll need to run a local postgres DB and use it to refresh the migration and generate entities (as explained in https://github.com/RGB-Tools/rgb-lib/blob/master/migration/README.md). At this point if you'll look to the changes (`git diff`) you'll see changes to several files, but you should keep (`git add -p`) only the changes that are relevant to the change you made to the init file (i.e. adding the new `ExactExpiry` field). For example in this PR you pushed changes like:
```diff
-    pub status: TransferStatus,
+    pub status: i16,
```
or 
```diff
-    pub attachment_id: Option<u8>,
+    pub attachment_id: Option<i16>,
```
but these are not changes related to the addition of the new boolean field.
Look at the diff of this PR, the only **entity** file that should have changes is `src/database/entities/batch_transfer.rs` and it should have only the changes related to the addition of the new `ExactExpiry` field (a line in the `Model`, one in the `Column` and one in the `match` inside the `ColumnTrait` implementation)

PR #74, opened 2026-04-17
add support for P2WPKH

dcorral:
# PR: Add P2WPKH (SegWit V0) support

## Summary

- Add `ScriptType` enum (`P2wpkh` / `P2tr`) with `P2tr` as the default, preserving full backward compatibility
- Parameterize key derivation and descriptor generation to support both BIP84 (purpose 84) and BIP86 (purpose 86) paths
- Add `script_type` field to `WalletData` with `#[serde(default)]` so existing serialized data continues to work
- Add `generate_keys_with_script_type` and `restore_keys_with_script_type` public API functions
- Add integration tests for P2WPKH-only and cross-type (P2TR to P2WPKH) RGB transfers
- Update UniFFI bindings with `ScriptType` enum and `WalletData.script_type` field

## Backward compatibility

- `ScriptType` defaults to `P2tr` — existing wallets and serialized `WalletData` are unaffected
- `generate_keys` / `restore_keys` unchanged in signature and behavior
- `#[serde(default)]` on `script_type` means JSON without the field deserializes to `P2tr`

## Test check

- [x] `cargo test -- segwit_v0` — both P2WPKH integration tests passed

### DIFF
--- a/bindings/uniffi/src/rgb-lib.udl
+++ b/bindings/uniffi/src/rgb-lib.udl
@@ -1,8 +1,8 @@
 namespace rgb_lib {
-  Keys generate_keys(BitcoinNetwork bitcoin_network);
+  Keys generate_keys(BitcoinNetwork bitcoin_network, WitnessVersion witness_version);
 
   [Throws=RgbLibError]
-  Keys restore_keys(BitcoinNetwork bitcoin_network, string mnemonic);
+  Keys restore_keys(BitcoinNetwork bitcoin_network, string mnemonic, WitnessVersion witness_version);
 
   [Throws=RgbLibError]
   void restore_backup(string backup_path, string password, string data_dir);
@@ -71,6 +71,7 @@ interface RgbLibError {
   InvalidTransportEndpoints(string details);
   InvalidTxid();
   InvalidVanillaKeychain();
+  InvalidWitnessVersion(string witness_version);
   MaxFeeExceeded(string txid);
   MinFeeNotMet(string txid);
   MultisigHubService(string details);
@@ -271,6 +272,12 @@ enum DatabaseType {
   "Sqlite",
 };
 
+[Remote]
+enum WitnessVersion {
+  "SegWitV0",
+  "Taproot",
+};
+
 interface Address {
   [Throws=RgbLibError]
   constructor(string address_string, BitcoinNetwork bitcoin_network);
@@ -330,6 +337,7 @@ dictionary Keys {
   string account_xpub_vanilla;
   string account_xpub_colored;
   string master_fingerprint;
+  WitnessVersion witness_version;
 };
 
 [Remote]
@@ -525,6 +533,7 @@ dictionary SinglesigKeys {
   u8? vanilla_keychain;
   string master_fingerprint;
   string? mnemonic;
+  WitnessVersion witness_version;
 };
 
 [Remote]
--- a/src/error.rs
+++ b/src/error.rs
@@ -373,6 +373,13 @@ pub enum Error {
     #[error("Invalid vanilla keychain")]
     InvalidVanillaKeychain,
 
+    /// Invalid witness version
+    #[error("Invalid witness version: {witness_version}")]
+    InvalidWitnessVersion {
+        /// The invalid witness version
+        witness_version: String,
+    },
+
     /// The maximum fee has been exceeded
     #[error("Max fee exceeded for transfer with TXID: {txid}")]
     MaxFeeExceeded {
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -4,6 +4,54 @@
 
 use super::*;
 
+/// Supported bitcoin witness versions.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub enum WitnessVersion {
+    /// Segregated Witness version 0
+    SegWitV0,
+    /// Segregated Witness version 1
+    #[default]
+    Taproot,
+}
+
+impl WitnessVersion {
+    pub(crate) fn purpose(&self) -> u32 {
+        match self {
+            WitnessVersion::SegWitV0 => 84,
+            WitnessVersion::Taproot => 86,
+        }
+    }
+
+    pub(crate) fn descriptor_fn(&self) -> &'static str {
+        match self {
+            WitnessVersion::SegWitV0 => "wpkh",
+            WitnessVersion::Taproot => "tr",
+        }
+    }
+}
+
+impl fmt::Display for WitnessVersion {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+impl FromStr for WitnessVersion {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s.to_lowercase().as_str() {
+            "segwitv0" => WitnessVersion::SegWitV0,
+            "taproot" => WitnessVersion::Taproot,
+            _ => {
+                return Err(Error::InvalidWitnessVersion {
+                    witness_version: s.to_string(),
+                });
+            }
+        })
+    }
+}
+
 /// A set of Bitcoin keys used by the wallet.
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
@@ -18,10 +66,13 @@ pub struct Keys {
     pub account_xpub_colored: String,
     /// Fingerprint of the master xPub
     pub master_fingerprint: String,
+    /// Witness version these keys were derived with
+    #[serde(default)]
+    pub witness_version: WitnessVersion,
 }
 
-/// Generate a set of [`Keys`] for the given Bitcoin network.
-pub fn generate_keys(bitcoin_network: BitcoinNetwork) -> Keys {
+/// Generate a set of [`Keys`] for the given Bitcoin network and witness version.
+pub fn generate_keys(bitcoin_network: BitcoinNetwork, witness_version: WitnessVersion) -> Keys {
     let bdk_network = BdkNetwork::from(bitcoin_network);
     let mnemonic = Mnemonic::generate((WordCount::Words12, Language::English))
         .expect("to be able to generate a new mnemonic");
@@ -32,7 +83,7 @@ pub fn generate_keys(bitcoin_network: BitcoinNetwork) -> Keys {
     let xpub = &xkey.into_xpub(bdk_network, &Secp256k1::new());
     let mnemonic_str = mnemonic.to_string();
     let (account_xpub_vanilla, account_xpub_colored) =
-        get_account_xpubs(&bitcoin_network, &mnemonic_str).unwrap();
+        get_account_xpubs(&bitcoin_network, &mnemonic_str, witness_version).unwrap();
     let master_fingerprint = xpub.fingerprint().to_string();
     Keys {
         mnemonic: mnemonic_str,
@@ -40,14 +91,19 @@ pub fn generate_keys(bitcoin_network: BitcoinNetwork) -> Keys {
         account_xpub_vanilla: account_xpub_vanilla.to_string(),
         account_xpub_colored: account_xpub_colored.to_string(),
         master_fingerprint,
+        witness_version,
     }
 }
 
-/// Recreate a set of [`Keys`] from the given mnemonic phrase.
-pub fn restore_keys(bitcoin_network: BitcoinNetwork, mnemonic: String) -> Result<Keys, Error> {
+/// Recreate a set of [`Keys`] from the given mnemonic phrase for the given witness version.
+pub fn restore_keys(
+    bitcoin_network: BitcoinNetwork,
+    mnemonic: String,
+    witness_version: WitnessVersion,
+) -> Result<Keys, Error> {
     let bdk_network = BdkNetwork::from(bitcoin_network);
     let (account_xpub_vanilla, account_xpub_colored) =
-        get_account_xpubs(&bitcoin_network, &mnemonic)?;
+        get_account_xpubs(&bitcoin_network, &mnemonic, witness_version)?;
     let mnemonic_parsed = Mnemonic::parse_in(Language::English, &mnemonic)?;
     let xkey: ExtendedKey = mnemonic_parsed
         .clone()
@@ -61,12 +117,30 @@ pub fn restore_keys(bitcoin_network: BitcoinNetwork, mnemonic: String) -> Result
         account_xpub_vanilla: account_xpub_vanilla.to_string(),
         account_xpub_colored: account_xpub_colored.to_string(),
         master_fingerprint,
+        witness_version,
     })
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
+
+    #[test]
+    fn witness_version_display_and_parse() {
+        for wv in [WitnessVersion::SegWitV0, WitnessVersion::Taproot] {
+            let s = wv.to_string();
+            assert_eq!(WitnessVersion::from_str(&s).unwrap(), wv);
+            assert_eq!(WitnessVersion::from_str(&s.to_lowercase()).unwrap(), wv);
+        }
+
+        let err = WitnessVersion::from_str("nonsense").unwrap_err();
+        assert_eq!(
+            err,
+            Error::InvalidWitnessVersion {
+                witness_version: "nonsense".to_string(),
+            },
+        );
+    }
 
     #[test]
     fn generate_success() {
@@ -76,7 +150,8 @@ mod test {
             account_xpub_vanilla,
             account_xpub_colored,
             master_fingerprint,
-        } = generate_keys(BitcoinNetwork::Regtest);
+            witness_version,
+        } = generate_keys(BitcoinNetwork::Regtest, WitnessVersion::Taproot);
 
         assert!(Mnemonic::from_str(&mnemonic).is_ok());
         let pubkey = Xpub::from_str(&xpub);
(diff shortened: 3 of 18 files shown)


src/wallet/objects.rs
```diff
@@ -31,6 +59,9 @@ pub struct WalletData {
     /// List of schemas the wallet should support (when issuing, sending and receiving). Empty list
     /// means the wallet should support all the schemas rgb-lib supports.
     pub supported_schemas: Vec<AssetSchema>,
+    /// Script type for wallet addresses. Default: P2TR (taproot).
+    #[serde(default)]
+    pub script_type: ScriptType,
```
zoedberg: Adding the script type to the `WalletData` seems wrong. Since the script type is a property of the key derivation it should be added to `Keys` and `SinglesigKeys` (and `MultisigKeys`), otherwise we open the door to issues like keys created with a script type and used on a wallet specifying a different script type.

src/keys.rs
```diff
@@ -21,7 +21,17 @@ pub struct Keys {
...
 
 /// Generate a set of [`Keys`] for the given Bitcoin network.
+///
+/// Uses P2TR (taproot, purpose 86) derivation paths by default.
 pub fn generate_keys(bitcoin_network: BitcoinNetwork) -> Keys {
+    generate_keys_with_script_type(bitcoin_network, ScriptType::default())
+}
+
+/// Generate a set of [`Keys`] for the given Bitcoin network and script type.
+pub fn generate_keys_with_script_type(
```
zoedberg: In this phase we don't worry about changes to public methods, so please keep only the `generate_keys` and `restore_keys` methods, adding the `script_type` to those

src/wallet/test/mod.rs
```diff
@@ -331,6 +331,7 @@ mod multisig;
 mod new;
 mod refresh;
 mod rust_only;
+mod segwit_v0;
```
zoedberg: Please rename this file, saying segwit v0 is imprecise. This PR adds support for P2WPKH, not for all segwit v0 scripts

zoedberg: Please rename this file, saying segwit v0 is imprecise. This PR adds support for P2WPKH, not for all segwit v0 scripts

src/wallet/test/segwit_v0.rs
```diff
@@ -0,0 +1,62 @@
+use super::*;
+
+// End-to-end RGB on P2WPKH: fund + create_utxos + issue NIA + blind_receive + send.
+// Exercises the OpretFirst commitment on a non-taproot output.
```
zoedberg: it probably makes more sense to add tests to existing test modules of APIs that we think could be affected by the different descriptor instead of having this dedicated mod

src/utils.rs
```diff
@@ -300,24 +301,40 @@ fn get_xpub_from_xprv(xprv: &Xpriv) -> Xpub {
...
+) -> Result<(Xpriv, Xpub, Fingerprint), Error> {
+    get_account_data_for_purpose(bitcoin_network, mnemonic, rgb, PURPOSE)
+}
+
+/// Like [`get_account_data`], but parameterised on the BIP purpose (84 for P2WPKH, 86 for P2TR).
+pub fn get_account_data_for_purpose(
+    bitcoin_network: &BitcoinNetwork,
+    mnemonic: &str,
+    rgb: bool,
+    purpose: u8,
```
zoedberg: Let's change `get_account_data` instead of adding a new method. Also do not take purpose as u8, get the script type and only when we need to extract the purpose let's convert the script type to purpose

zoedberg (changes requested):
Please see the requested changes. Also please rename the commit since `Segwit_v0 support` is imprecise, this PR is adding support for just P2WPKH, not for all segwit v0 scripts. I would rename it to `add support for P2WPKH`

src/wallet/objects.rs
```diff
@@ -15,6 +15,56 @@ pub enum DatabaseType {
     Sqlite,
 }
 
+/// Supported script types for wallet addresses.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub enum ScriptType {
```
zoedberg: I would move this to the `keys` module

src/wallet/objects.rs
```diff
@@ -15,6 +15,56 @@ pub enum DatabaseType {
     Sqlite,
 }
 
+/// Supported script types for wallet addresses.
```
```suggestion
/// Supported bitcoin script types.
```

src/wallet/objects.rs
```diff
@@ -15,6 +15,56 @@ pub enum DatabaseType {
...
+    /// Pay-to-Witness-Public-Key-Hash (SegWit V0, BIP84)
+    P2wpkh,
+    /// Pay-to-Taproot (SegWit V1, BIP86)
+    #[default]
+    P2tr,
+}
+
+impl ScriptType {
+    /// BIP purpose number for this script type.
+    pub fn purpose(&self) -> u8 {
```
```suggestion
    pub(crate) fn purpose(&self) -> u8 {
```
I think this method can stay private

src/wallet/objects.rs
```diff
@@ -15,6 +15,56 @@ pub enum DatabaseType {
...
+        match self {
+            ScriptType::P2wpkh => 84,
+            ScriptType::P2tr => 86,
+        }
+    }
+
+    /// Descriptor format prefix for this script type.
```
```suggestion
```
We don't define docstrings on private methods. If a private method is not self-documenting then we add a comment, but in this case I think it's not needed

src/error.rs
```diff
@@ -190,6 +190,13 @@ pub enum Error {
         network: String,
     },
 
+    /// Invalid script type
+    #[error("Invalid script type: {script_type}")]
+    InvalidScriptType {
```
zoedberg: Please respect the alphabetical order

src/utils.rs
```diff
@@ -272,9 +272,9 @@ pub(crate) fn get_coin_type(bitcoin_network: &BitcoinNetwork, rgb: bool) -> u32
     }
 }
 
-pub(crate) fn get_account_derivation_children(coin_type: u32) -> Vec<ChildNumber> {
+pub(crate) fn get_account_derivation_children(purpose: u8, coin_type: u32) -> Vec<ChildNumber> {
```
zoedberg: Please get `ScriptType` here. This way instead of calling the `purpose()` method in multiple places you call it only once, where it's actually needed. Also, I would change the return type of the `purpose()` method to return a u32, to avoid conversion here

src/wallet/test/multisig/mod.rs
```diff
@@ -1262,3 +1262,36 @@ fn fail() {
...
+        Cosigner::from_keys(&tr_keys_2, None),
+        Cosigner::from_keys(&wpkh_keys, None),
+    ];
+    let multisig_wlt_keys = MultisigKeys::new(cosigners, 2, 2);
+
+    let res = MultisigWallet::new(get_test_wallet_data(&data_dir), multisig_wlt_keys);
+    assert_matches!(
+        res.as_ref().err().unwrap(),
+        Error::InvalidCosigner { details }
+            if details.contains("P2tr") && details.contains("P2wpkh")
```
zoedberg: why don't you check the full details message here?

src/wallet/multisig.rs
```diff
@@ -115,6 +120,19 @@ impl MultisigKeys {
...
         }
+        // multisig is only supported for P2TR keys: the descriptor template is
+        // tr(NUMS, multi_a(...)) and cosigner xpubs are expected at purpose 86
+        for c in &self.cosigners {
+            if c.script_type != ScriptType::P2tr {
+                return Err(Error::InvalidCosigner {
+                    details: format!(
+                        "multisig wallets only support {:?} keys, got {:?}",
+                        ScriptType::P2tr,
+                        c.script_type
```
zoedberg: I would just say "unsupported script type"

src/wallet/multisig.rs
```diff
@@ -115,6 +120,19 @@ impl MultisigKeys {
         if self.cosigners.is_empty() {
             return Err(Error::NoCosignersSupplied);
         }
+        // multisig is only supported for P2TR keys: the descriptor template is
+        // tr(NUMS, multi_a(...)) and cosigner xpubs are expected at purpose 86
```
zoedberg:
```suggestion
        // multisig wallets support only P2TR for now
```

src/wallet/multisig.rs
```diff
@@ -115,6 +120,19 @@ impl MultisigKeys {
...
+        for c in &self.cosigners {
```
zoedberg: Instead of looping twice on the cosigners can't we do this check together with the other cosigner checks?

src/wallet/multisig.rs
```diff
@@ -24,6 +24,9 @@ pub struct Cosigner {
     pub vanilla_keychain: Option<u8>,
     /// Master fingerprint
     pub master_fingerprint: String,
+    /// Script type these xpubs were derived for
+    #[serde(default)]
+    pub script_type: ScriptType,
```
zoedberg: I'm not sure it's possible to have mixed cosigners (i.e. with different script types), is it? If not please remove this from here

src/wallet/singlesig.rs
```diff
@@ -19,6 +19,9 @@ pub struct SinglesigKeys {
     pub master_fingerprint: String,
     /// Wallet mnemonic phrase
     pub mnemonic: Option<String>,
+    /// Script type these keys were derived for
```
zoedberg:
```suggestion
    /// Script type these keys were derived with
```

src/keys.rs
```diff
@@ -18,10 +18,13 @@ pub struct Keys {
     pub account_xpub_colored: String,
     /// Fingerprint of the master xPub
     pub master_fingerprint: String,
+    /// Script type these keys were derived for
```
zoedberg:
```suggestion
    /// Script type these keys were derived with
```

src/keys.rs
```diff
@@ -100,12 +111,45 @@ mod test {
...
+        assert_eq!(keys_tr.xpub, keys_wpkh.xpub);
+        assert_eq!(keys_tr.master_fingerprint, keys_wpkh.master_fingerprint);
+        // but different account xpubs (different purpose in derivation path)
+        assert_ne!(keys_tr.account_xpub_colored, keys_wpkh.account_xpub_colored);
+        assert_ne!(keys_tr.account_xpub_vanilla, keys_wpkh.account_xpub_vanilla);
+        // and script_type is propagated
+        assert_eq!(keys_tr.script_type, ScriptType::P2tr);
+        assert_eq!(keys_wpkh.script_type, ScriptType::P2wpkh);
+
+        // restore round-trip for P2WPKH
```
zoedberg: let's move this to `restore_success`

zoedberg (changes requested):
Please address the requested changes

src/wallet/multisig.rs
dcorral: Agree, I promoted Cosigner::from_keys to a fallible constructor that rejects non-P2TR keys with Error::InvalidCosigner { details: "unsupported script type" }.

Open question: should the script type live on MultisigKeys instead (single field, MultisigKeys::new validates == P2tr)? 
This way it documents intent, fail-fast on deserialised JSON, natural spot to relax later when implementation for other types come to multisig. Since we only support tr now it's redundant today though. What are your thoghts on this?

zoedberg: The `Cosigner::from_keys` is a test method, so we don't need to change that until we add support for segwit v0 also there.

When the script type was defined at the `WalletData` level (object used by both single and multi sig) it made sense to check that the multisig wallet couldn't be initiated with p2wpkh, but now that you moved it to the `Keys` is actually not necessary. But since I'll probably soon add support for multisig too (which will be p2wsh instead of p2wpkh), let's use a more generic enum:
```
enum WitnessVersion {
    SegWitV0,
    Taproot,
}
```

src/wallet/test/send.rs
```diff
@@ -7822,3 +7822,64 @@ fn allocations() {
...
+fn p2wpkh_send_receive_nia() {
+    initialize();
+
+    let amount: u64 = 66;
+
+    let (mut wallet, online) = get_funded_wallet_p2wpkh();
+    let (mut rcv_wallet, _rcv_online) = get_funded_wallet_p2wpkh();
+
+    let asset = test_issue_asset_nia(&mut wallet, online, None);
+    let receive_data = test_blind_receive(&mut rcv_wallet);
```
zoedberg: let's also add a witness receive and send to both recipients

src/wallet/test/send.rs
```diff
@@ -7822,3 +7822,64 @@ fn allocations() {
...
+    let recipient_map = HashMap::from([(
+        asset.asset_id.clone(),
+        vec![Recipient {
+            assignment: Assignment::Fungible(amount),
+            recipient_id: receive_data.recipient_id.clone(),
+            witness_data: None,
+            transport_endpoints: TRANSPORT_ENDPOINTS.clone(),
+        }],
+    )]);
+    let txid = test_send(&mut wallet, online, &recipient_map);
```
zoedberg, line 7867: test is not complete without a refresh and a settled balance check (on both sides)

src/wallet/test/send.rs
```diff
@@ -7822,3 +7822,64 @@ fn allocations() {
...
```
zoedberg, line 7910: also here the send is not complete without the refreshes. please also send some assets back so that we have the full roundtrip

src/wallet/mod.rs
```diff
@@ -17,6 +17,7 @@ pub(crate) mod singlesig;
 #[cfg(test)]
 pub(crate) mod test;
 
+pub use crate::keys::ScriptType;
```
zoedberg: the `keys` module is already public, users can import this from there, no need to re-export it

src/keys.rs
```diff
@@ -4,6 +4,55 @@
...
+pub enum ScriptType {
+    /// Pay-to-Witness-Public-Key-Hash (SegWit V0, BIP84)
+    P2wpkh,
+    /// Pay-to-Taproot (SegWit V1, BIP86)
+    #[default]
+    P2tr,
+}
+
+impl ScriptType {
+    /// BIP purpose number for this script type.
```
zoedberg:
```suggestion
```
no docstrings on private methods please

src/keys.rs
```diff
@@ -4,6 +4,55 @@
...
+            "p2wpkh" => ScriptType::P2wpkh,
+            "p2tr" => ScriptType::P2tr,
+            _ => {
+                return Err(Error::InvalidScriptType {
+                    script_type: s.to_string(),
+                });
+            }
+        })
+    }
+}
```
zoedberg: we need a unit test for roundtrip string conversion

zoedberg (changes requested):
some other requests

bindings/uniffi/src/rgb-lib.udl
```diff
@@ -66,6 +66,7 @@ interface RgbLibError {
   InvalidRecipientMap();
   InvalidRecipientNetwork();
   InvalidRejectListUrl(string details);
+  InvalidWitnessVersion(string witness_version);
```
zoedberg: Please respect the alphabetical order

src/keys.rs
```diff
@@ -4,6 +4,54 @@
...
 
+/// Supported bitcoin witness versions.
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub enum WitnessVersion {
+    /// Segregated Witness version 0 (P2WPKH / P2WSH, BIP84 / BIP48)
+    SegWitV0,
+    /// Segregated Witness version 1 (Taproot, BIP86)
+    #[default]
+    Taproot,
+}
```
zoedberg, line 15: Let’s remove the parentheses to keep the docstrings simple.
The BIPs referenced here mostly relate to wallet derivation conventions rather than the witness version consensus rules.

src/lib.rs
```diff
@@ -40,12 +40,13 @@
 //! ### Create an RGB singlesig wallet
 //! ```
 //! use rgb_lib::keys::generate_keys;
+//! use rgb_lib::keys::WitnessVersion;
```
zoedberg: Please group this with the other import from the keys module

src/lib.rs
```diff
@@ -83,6 +84,7 @@ pub use rgbstd::{
 pub use crate::{
     database::enums::{AssetSchema, Assignment, TransferStatus, TransportType},
     error::Error,
+    keys::WitnessVersion,
```
zoedberg: Please do not re-export this, they keys module is already public

src/utils.rs
```diff
@@ -299,25 +303,29 @@ fn get_xpub_from_xprv(xprv: &Xpriv) -> Xpub {
     Xpub::from_priv(&Secp256k1::new(), xprv)
 }
 
-/// Get the account-level xPriv and xPub for the given mnemonic and Bitcoin network based on the
-/// requested wallet side (colored or vanilla)
+/// Get the account-level xPriv and xPub for the given mnemonic, Bitcoin network and script type,
```
zoedberg: This needs to be fixed after the rename from script type to witness version

src/wallet/test/send.rs
```diff
@@ -255,10 +255,10 @@ fn success() {
     assert!(consignment.result.is_some());
     // settle transfer
     wait_for_refresh(&mut rcv_wallet, rcv_online, None, None);
-    wait_for_refresh(&mut wallet, online, Some(&asset.asset_id), None);
+    wait_for_refresh(&mut wallet, online, None, None);
     mine(false, false);
     wait_for_refresh(&mut rcv_wallet, rcv_online, None, None);
-    wait_for_refresh(&mut wallet, online, Some(&asset.asset_id), None);
+    wait_for_refresh(&mut wallet, online, None, None);
```
zoedberg: Please do not make unnecessary changes to existing tests, here and in all other points where you made such changes

zoedberg (changes requested):
Please address the requested changes

zoedberg (approved):
LGTM, thank you!

PR #75, opened 2026-04-21
Add derivation index to unspent list

berserker-systems:
This method adds a way to prove the ownership of an asset. We need it to allow access to a service related to an asset.
If there is anything to improve, please let me know. The message is there to prevent signature-reuse.

### DIFF
--- a/src/wallet/objects.rs
+++ b/src/wallet/objects.rs
@@ -228,6 +228,24 @@ impl From<RgbEmbeddedMedia> for EmbeddedMedia {
     }
 }
 
+/// A cryptographic proof of UTXO ownership via message signing.
+///
+/// Contains a Bitcoin message signature and the public key that produced it,
+/// allowing a third party to verify that the signer controls the private key
+/// corresponding to the UTXO's scriptPubKey.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
+pub struct UtxoSignature {
+    /// The outpoint (txid:vout) of the UTXO whose ownership is being proved
+    pub outpoint: Outpoint,
+    /// The message that was signed (the SHA256 digest, not the raw input)
+    pub message: Vec<u8>,
+    /// The BIP-340 Schnorr signature (64 bytes)
+    pub signature: Vec<u8>,
+    /// The 32-byte x-only taproot-tweaked public key matching the P2TR output's scriptPubKey
+    pub pubkey: Vec<u8>,
+}
+
 /// A proof of reserves.
 #[derive(Debug, Clone, Default, PartialEq, Eq, Deserialize, Serialize)]
 #[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
--- a/src/wallet/objects.rs
+++ b/src/wallet/objects.rs
@@ -1370,6 +1370,10 @@ pub struct Utxo {
     pub colorable: bool,
     /// Defines if the UTXO already exists (TX that creates it has been broadcasted)
     pub exists: bool,
+    /// Derivation index of the UTXO's script pubkey. `None` if the transaction creating the UTXO
+    /// is not known to the underlying BDK wallet, which is the case while the UTXO does not yet
+    /// exist on-chain and for outputs that do not belong to this wallet.
+    pub derivation_index: Option<u32>,
 }
 
 impl From<DbTxo> for Utxo {
@@ -1382,6 +1386,7 @@ impl From<DbTxo> for Utxo {
                 .expect("DB should contain a valid u64 value"),
             colorable: true,
             exists: x.exists,
+            derivation_index: None,
         }
     }
 }
@@ -1393,6 +1398,7 @@ impl From<LocalOutput> for Utxo {
             btc_amount: x.txout.value.to_sat(),
             colorable: false,
             exists: true,
+            derivation_index: Some(x.derivation_index),
         }
     }
 }
--- a/src/wallet/offline.rs
+++ b/src/wallet/offline.rs
@@ -2188,6 +2188,14 @@ pub trait WalletOffline: WalletBackup {
                 .for_each(|u| u.rgb_allocations.retain(|a| a.settled));
         }
 
+        let spk_index = self.bdk_wallet().spk_index();
+        for unspent in unspents.iter_mut() {
+            if let Some(((KeychainKind::External, derivation_index), _)) =
+                spk_index.txout(BdkOutPoint::from(unspent.utxo.outpoint.clone()))
+            {
+                unspent.utxo.derivation_index = Some(derivation_index);
+            }
+        }
         let mut internal_unspents: Vec<Unspent> =
             self.internal_unspents().map(Unspent::from).collect();
 
--- a/src/wallet/test/list_unspents.rs
+++ b/src/wallet/test/list_unspents.rs
@@ -1,6 +1,33 @@
 use super::*;
 
 #[cfg(feature = "electrum")]
+fn check_unspents_derivation(wallet: &Wallet, unspents: &[Unspent]) {
+    for unspent in unspents.iter().filter(|u| u.utxo.exists) {
+        let outpoint = &unspent.utxo.outpoint;
+        let Some(derivation_index) = unspent.utxo.derivation_index else {
+            panic!("existing unspent {outpoint} is missing derivation index");
+        };
+        let keychain = if unspent.utxo.colorable {
+            KeychainKind::External
+        } else {
+            KeychainKind::Internal
+        };
+        let derived_spk = wallet
+            .bdk_wallet()
+            .public_descriptor(keychain)
+            .at_derivation_index(derivation_index)
+            .unwrap()
+            .script_pubkey();
+        let txid: bdk_wallet::bitcoin::Txid = unspent.utxo.outpoint.txid.parse().unwrap();
+        let tx = wallet.bdk_wallet().get_tx(txid).unwrap();
+        let actual_spk = tx.tx_node.tx.output[unspent.utxo.outpoint.vout as usize]
+            .script_pubkey
+            .clone();
+        assert_eq!(derived_spk, actual_spk);
+    }
+}
+
+#[cfg(feature = "electrum")]
 #[test]
 #[parallel]
 fn success() {
@@ -31,6 +58,12 @@ fn success() {
     let unspent_list_all = party.list_unspents(false);
     assert_eq!(unspent_list_all.len(), 1);
     assert!(unspent_list_all.iter().all(|u| !u.utxo.colorable));
+    assert!(
+        unspent_list_all
+            .iter()
+            .all(|u| u.utxo.derivation_index.is_some())
+    );
+    check_unspents_derivation(&party.wallet, &unspent_list_all);
 
     party.create_utxos_default();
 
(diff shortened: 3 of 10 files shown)


zoedberg:
The code has several issues, both on implementation, like usage of `unwrap` and missing logic to handle `SegwitV0` wallets (recently added) and on logic, like missing support for assets allocated on blinded UTXOs.

But before discussing these issues I would like to understand if this method is actually needed. From what I see, everything can already be done with what rgb-lib provides, just call `list_unspents` to get the list of outpoints where the assets are, select the outpoint for which you want to prove ownership and then use whatever bitcoin library and language you want to sign this information.

berserker-systems:
@zoedberg 
I agree with your observations. SegwitV0 needs to be added and to support blinded UTXOs the interface needs to change.                                                

 > everything can already be done with what rgb-lib provides                                                                                                            
                   
The caller would need to duplicate the BDK wallet from the same descriptors, correct BIP-32 tree, witness version, keychain index, network-aware seed derivation. I'd rather not push that duplication onto callers.

More importantly, signing an outpoint isn't a free operation outside the library. BDK's derivation_of_spk only knows about indices it has indexed, so the duplicated wallet has to be synced against a chain source first (or the caller has to manually reveal addresses up to the right index) before the scriptPubKey -> (keychain, index) lookup works. Otherwise it returns None and the only way to recover the index is brute-forcing the child range until a derivation matches. rgb-lib already has a synced wallet with this state populated, so doing it inside is essentially free. Doing it outside means a full sync just to sign a message.

zoedberg:
BDK has a way to sync a single SPK, I think you could use that to avoid syncing a lot of unnecessary SPKs. With that you can do this operation outside rgb-lib without a noticeable overhead.

berserker-systems:
Single-SPK sync helps with chain sync, but the bottleneck is the lookup itself.                                                                                        

Given an outpoint from list_unspents, the caller doesn't know the (keychain, derivation_index) pair. To find it they have to derive children 0..N on the colored keychain, apply the tap-tweak (or not, depending on witness version), and compare each derived SPK against the outpoint's scriptPubKey until one matches. Single-SPK sync doesn't shortcut that, it's a derivation problem, not a network one. 

The state that does shortcut it is BDK's reverse SPK -> (keychain, index) map, which only contains entries for already-revealed indices. rgb-lib's wallet has this populated as part of normal operation; an external caller starting from a fresh duplicated wallet has to either rebuild it (= reveal/derive up to N) or brute-force on every signing.

Plus, to derive at all the caller needs the descriptors and the witness version, wallet-internal conventions they'd have to read out of SinglesigKeys and apply correctly.


And there's the README:

> This library is intended to exclusively handle all UTXOs for the wallet. Using the same mnemonic phrase on any other device, including with this same library, can lead to serious issues and ultimately to RGB asset loss.

I realise the strict reading of that is about operating on UTXOs from a second wallet, not read-only key derivation, so signing without broadcasting is technically safe. But it's still using the mnemonic somewhere other than rgb-lib, and I'd really rather not push callers toward that pattern, even for "harmless" signing operations, it's the kind of thing that drifts over time.

zoedberg:
What if we add the keychain and derivation index information to the unspent list?

berserker-systems:
@zoedberg 
Adding keychain and derivation index to the unspent list would solve the reverse lookup problem, so I agree that it would make external signing more practical.

To better understand the preferred API boundary: what is the main reason you would not want rgb-lib itself to provide a controlled signing/proof method for wallet-owned UTXOs? Is the concern mostly about keeping rgb-lib’s API surface and maintenance burden minimal, avoiding responsibility for generic Bitcoin signing operations, or something else?

From my perspective, proving ownership of an RGB asset feels like something that naturally belongs in an RGB library, especially if rgb-lib already owns the wallet state and key-derivation context needed to do it safely and consistently. I may be biased here since this is my use case, but that is why I initially expected this functionality to fit within rgb-lib. That said, I am open to changing the API/interface if there is a better shape for this functionality or if a more general signing/proof abstraction would fit the library better.

nicbus:
The main concern is keeping the rgb-lib's public APIs clear and maintainable.

Adding a new API to provide functionality for a specific use case is outside the scope of the project, which aims to provide an abstraction over RGB primitives that is as simple as possible while retaining the needed flexibility. Adding info to an existing API which enables more use cases fits with the goals and doesn't impact maintenance much.

As for proving ownership of UTXOs (not RGB assets specifically), it doesn't look like something that adds value to rgb-lib, as for all current intents and purposes, ownership is already proven when spending and there are no cases where proving control over a UTXO seems useful on its own.

berserker-systems:
@zoedberg @nicbus 
As discussed, I changed the PR to enrich the unspent list.

src/wallet/offline.rs
```diff
@@ -2175,10 +2175,28 @@ pub trait WalletOffline: WalletBackup {
...
-            self.internal_unspents().map(Unspent::from).collect();
-
-        unspents.append(&mut internal_unspents);
+        let mut to_enrich: HashMap<BdkOutPoint, usize> = unspents
+            .iter()
+            .enumerate()
+            .map(|(i, u)| (BdkOutPoint::from(u.utxo.outpoint.clone()), i))
+            .collect();
+        let mut vanilla_unspents: Vec<Unspent> = vec![];
+        for output in self.bdk_wallet().list_output() {
```
zoedberg: `list_output()` iterates all spent + unspent outputs; that feels heavy for every `list_unspents` call. Could we enrich colored UTXOs with `get_tx` + `derivation_of_spk` and keep `internal_unspents()` for vanilla? That should also handle almost-spent colored UTXOs without a full graph walk.

src/wallet/test/list_unspents.rs
```diff
@@ -1,5 +1,27 @@
...
+fn check_unspents_derivation(wallet: &Wallet, unspents: &[Unspent]) {
+    for unspent in unspents.iter().filter(|u| u.utxo.exists) {
+        let keychain = match unspent.utxo.keychain.unwrap() {
+            UtxoKeychain::Colored => KeychainKind::External,
+            UtxoKeychain::Vanilla => KeychainKind::Internal,
+        };
+        let derived_spk = wallet
+            .bdk_wallet()
+            .public_descriptor(keychain)
+            .at_derivation_index(unspent.utxo.derivation_index.unwrap())
```
zoedberg: Tests currently only cover Taproot wallets. Would you mind adding a `SegWitV0` case so we're confident `at_derivation_index` / descriptor handling works for both witness versions?

zoedberg (changes requested):
Thanks! The implementation now matches what we discussed in the thread, I updated the PR title accordingly. I requested some changes but overall looks good to me

berserker-systems:
@zoedberg Thank you, review addressed. Please let me know if there is more left to do.

src/wallet/offline.rs
```diff
@@ -2175,10 +2175,25 @@ pub trait WalletOffline: WalletBackup {
...
+                );
+                unspent.utxo.keychain = Some(keychain.into());
+                unspent.utxo.derivation_index = Some(derivation_index);
+            }
+        }
+        let mut vanilla_unspents: Vec<Unspent> =
             self.internal_unspents().map(Unspent::from).collect();
 
-        unspents.append(&mut internal_unspents);
+        unspents.append(&mut vanilla_unspents);
```
zoedberg:
```suggestion
        unspents.append(&mut internal_unspents);
```
let's avoid unnecessary renames

src/wallet/objects.rs
```diff
@@ -1367,6 +1385,12 @@ pub struct Utxo {
     pub colorable: bool,
     /// Defines if the UTXO already exists (TX that creates it has been broadcasted)
     pub exists: bool,
+    /// Keychain of the UTXO. `None` when the UTXO is not (yet) part of the wallet's transaction
+    /// graph, e.g. when `exists` is false (pending witness receive or in-flight send change)
+    pub keychain: Option<UtxoKeychain>,
```
```suggestion
```
not needed, when `colorable=true` it's always UtxoKeychain:Colored and when `colorable=false` its always `UtxoKeychain::Vanilla` (drop the enum too)

src/wallet/offline.rs
```diff
@@ -2175,10 +2175,25 @@ pub trait WalletOffline: WalletBackup {
...
+            {
+                debug_assert_eq!(
+                    keychain,
+                    KeychainKind::External,
+                    "colorable unspent {} on the vanilla keychain",
+                    unspent.utxo.outpoint
+                );
+                unspent.utxo.keychain = Some(keychain.into());
+                unspent.utxo.derivation_index = Some(derivation_index);
+            }
```
zoedberg:
```suggestion
    if let Some(((KeychainKind::External, derivation_index), _)) =
        spk_index.txout(BdkOutPoint::from(unspent.utxo.outpoint.clone()))
    {
        unspent.utxo.derivation_index = Some(derivation_index);
    }
```
this way it's safe in release too

src/wallet/objects.rs
```diff
@@ -1367,6 +1385,12 @@ pub struct Utxo {
...
+    /// Derivation index of the UTXO's script pubkey. `None` in the same cases as
+    /// [`Utxo::keychain`]
```
zoedberg: Since we are dropping `keychain` this comment needs to be updated. I would update to "[...] `None` if the transaction creating the UTXO is not known to the underlying BDK wallet, which is the case while the UTXO does not yet exist on-chain and for outputs that do not belong to this wallet."

src/wallet/test/list_unspents.rs
```diff
@@ -222,6 +267,21 @@ fn success() {
...
+            .iter()
+            .filter(|u| !u.utxo.exists)
+            .all(|u| u.utxo.keychain.is_none() && u.utxo.derivation_index.is_none())
+    );
+    assert!(
+        unspent_list_all
+            .iter()
+            .filter(|u| u.utxo.exists)
+            .all(|u| u.utxo.keychain.is_some() && u.utxo.derivation_index.is_some())
+    );
```
zoedberg:
```suggestion
    // every existing UTXO has its derivation index set
    assert!(
        unspent_list_all
            .iter()
            .filter(|u| u.utxo.exists)
            .all(|u| u.utxo.derivation_index.is_some())
    );
```
the second assert says "when we claim to know, we do know, that's a real guarantee. the first was saying "when it's pending, we're ignorant", that's not a guarantee (it's just today's behavior)

src/wallet/test/list_unspents.rs
```diff
@@ -1,5 +1,33 @@
...
+        let outpoint = &unspent.utxo.outpoint;
+        let (Some(utxo_keychain), Some(derivation_index)) =
+            (unspent.utxo.keychain, unspent.utxo.derivation_index)
+        else {
+            panic!("existing unspent {outpoint} is missing keychain or derivation index");
+        };
+        let keychain = match utxo_keychain {
+            UtxoKeychain::Colored => KeychainKind::External,
+            UtxoKeychain::Vanilla => KeychainKind::Internal,
+        };
```
zoedberg:
```suggestion
        let keychain = if unspent.utxo.colorable { KeychainKind::External } else { KeychainKind::Internal };
```

zoedberg (changes requested):
Please address the requested changes and rebase on top of the updated `master` tip. @berserker-systems FYI I'm about to do a new release of rgb-lib, if you quickly update this PR this can be part of the next release

berserker-systems:
@zoedberg Thank you, I'm on it and push soon.

berserker-systems:
@zoedberg Done

zoedberg:
@berserker-systems I asked you to rebase this on the updated `master`, it seems you forgot to do it

berserker-systems:
@zoedberg Yep, I rebased against the wrong origin. Done

src/wallet/offline.rs
zoedberg: this hasn't been resolved though, still renaming a variable that doesn't need to be renamed

zoedberg (changes requested):
@berserker-systems It seems you haven't addressed a point

berserker-systems:
@zoedberg Done, I missed that one somehow

zoedberg (approved):
LGTM, thanks!

berserker-systems:
@zoedberg Great, thank you very much!

PR #82, opened 2026-06-12
fix: return error instead of panicking on asset change underflow

berserker-systems:


zoedberg:
Thanks for the PR.

I don’t think this test demonstrates a real bug. By the time the change amount is calculated, the code has already verified that the selected inputs are sufficient to cover the outputs.

The failing scenario only exists because the test uses a mock to alter the collected inputs after that point, which cannot happen in the real flow. If an error were to occur there in practice, it would likely indicate a more serious underlying issue, such as a corrupted database or inconsistent wallet state. In that case, we prefer the code to panic and stop execution rather than allow the wallet to continue operating and potentially worsen the situation.

For that reason, I’m going to close this PR.

berserker-systems:
@zoedberg 
Thank you for the info.
I ran in this issue with rgblib =0.3.0-beta.6
The wallet was fresh created and had less than 5 transactions. I'll have a look if the database was somehow inconsistent.

berserker-systems:
> we prefer the code to panic and stop execution rather than allow the wallet to continue operating and potentially worsen the situation

Short note: rust disables overflow checks in release builds by default. I do not see rgb-lib enabling `overflow-checks`.

zoedberg:
> @zoedberg Thank you for the info. I ran in this issue with rgblib =0.3.0-beta.6 The wallet was fresh created and had less than 5 transactions. I'll have a look if the database was somehow inconsistent.

Thanks for the follow-up.

Yes, please do investigate if you can. If this happened in production without mocks involved, then it likely means a more serious bug occurred earlier in the flow, leaving the wallet state inconsistent.

Any additional information would be very helpful. If you’re comfortable sharing it, the wallet directory, logs, exact steps that led to the issue, and any other related context could help us debug this further.

Thanks again for looking into it.

zoedberg:
> Short note: rust disables overflow checks in release builds by default. I do not see rgb-lib enabling overflow-checks.

Good point, thanks for noting that.

You’re right that Rust does not check integer overflows in release builds by default, and rgb-lib currently doesn't enable `overflow-checks`.

Instead of enabling it, I would rather check this explicitly, like:

```rust
let change_amount = input_amount
    .checked_sub(output_amount)
    .expect("selected inputs must cover outputs");
```

That said, in this specific flow the subtraction should still only happen after the code has verified that the selected inputs cover the outputs. So if this underflow is reachable in production, I would treat that as evidence that something earlier left the wallet state inconsistent, rather than as the root issue itself.

I’d still be very interested in debugging the production case you saw.

PR #83, opened 2026-06-15
fix: panic explicitly when RGB change would underflow

berserker-systems:


zoedberg:
Thank you! Please sign your commit though, in this repo merging is blocked if commits don't have verified signatures

berserker-systems:
@zoedberg 
Thank you, done!

PR #84, opened 2026-06-19
fixes and improvements (overflow-checks)

zoedberg:


src/error.rs
```diff
@@ -274,6 +274,10 @@ pub enum Error {
         details: String,
     },
 
+    /// The provided maximum number of allocations per UTXO is invalid
+    #[error("Invalid maximum allocations per UTXO: must be greater than 0")]
+    InvalidMaxAllocationsPerUtxo,
```
nicbus: given that the error is only about the value being zero, we should probably use a naming similar to e.g. `InvalidAmountZero` or `NoBurnAmount`/`NoInflationAmounts`.

I'd go with `NoMaxAllocationsPerUtxo`, documented as "Cannot allocate assets with zero allocations per UTXO"

src/wallet/test/create_utxos.rs
```diff
@@ -271,6 +271,32 @@ fn casting() {
...
+    party.create_utxos(false, Some(255), None, FEE_RATE, Some(255));
+    party.create_utxos(false, Some(1), None, FEE_RATE, Some(1));
+    let colorable = party
+        .list_unspents(false)
+        .iter()
+        .filter(|u| u.utxo.colorable)
+        .count();
+    assert_eq!(colorable, 256);
+
+    // 256 allocatable UTXOs exceed any up_to request, so no UTXO must be created
```
nicbus: It's not immediately clear to me what this test is supposed to trigger.
I'd add a add to the comment to clarify which regression this is preventing.

Proposal: "make sure casting to u8 is handled properly"

nicbus (changes requested):
a couple improvement proposals

src/wallet/test/create_utxos.rs
zoedberg: I’d rather keep the comment focused on the expected behavior than on one possible implementation issue. The regression is that when the wallet already has more allocatable UTXOs than the requested `up_to`, `create_utxos` must not create anything and should return `AllocationsAlreadyAvailable`. A bad u8 cast is only one way this can break, but the test is specifying the behavior.

I can change the comment to:
`256 allocatable UTXOs already exceed up_to=1, so creation must be rejected`
would it be more clear?

src/error.rs
zoedberg: done

src/wallet/test/create_utxos.rs
nicbus: What I find misleading about this approach is that the existing `up_to_allocation_checks` test already checks that `up_to` = `true` and `num` = `Some(1)`  returns `AllocationsAlreadyAvailable` when appropriate.

This test does the same but with more than `u8::max` available allocations, the cast is the main reason this exists and it's hard to guess outside the scope of this commit (and even inside, the link to the fix in src/wallet/online.rs is not obvious).

>  256 allocatable UTXOs already exceed up_to=1, so creation must be rejected

to me this is equivalent to the previous version

zoedberg: Fair point, the missing detail is the boundary being tested. I still prefer not to describe the test as only about a u8 cast, because the behavior we want to preserve is that `up_to` must compare against the actual number of available allocatable UTXOs, even when that number is above `u8::MAX`.

Maybe this is clearer:

`even with more than u8::MAX allocatable UTXOs, up_to=1 must detect that allocations are already available and reject creation`

nicbus: > the missing detail is the boundary being tested

yes, that describes well what I was trying to report

> even with more than u8::MAX allocatable UTXOs, up_to=1 must detect that allocations are already available and reject creation

it still doesn't hint at the connection but I guess that a small change could make it good enough:

when up_to is true, detect that the required num (u8) of allocations is already available and reject creation, even with more than u8::MAX already available allocations

zoedberg: done

nicbus (approved):

PR #87, opened 2026-06-26
Switch BDK persistence from file store to SQLite

markuslevonyak:
`rgb-lib` currently persists BDK wallet data with `bdk_file_store`, which upstream explicitly [documents](https://docs.rs/bdk_file_store/latest/bdk_file_store/) as unfit for production:

> `bdk_file_store` is a development/testing database. It does not natively support backwards compatible BDK version upgrades so should not be used in production.

This already caused breakage: the file store format changed in the BDK upgrade from 2.0 to 3.0, leaving existing databases unreadable. SQLite is the production-oriented, upgrade-migratable backend.

This PR moves BDK persistence to [`bdk_sqlite`](https://github.com/bitcoindevkit/bdk-sqlite) (BDK's `sqlx`-based SQLite store). Using it rather than `bdk_wallet`'s built-in `rusqlite` feature also avoids a native-library conflict: `rusqlite` and `rgb-lib`'s existing `sea-orm`/`sqlx` stack pull different `libsqlite3-sys` versions, and two crates linking `sqlite3` can't coexist. `bdk_sqlite` shares `sea-orm`'s `sqlx`, so the dependency graph resolves cleanly.

### Note on existing wallets

This switch (like the BDK 2.0 to 3.0 upgrade before it) is not backwards compatible: existing BDK file-store databases are not read, so on first open the wallet creates a fresh SQLite store and repopulates via the normal scan. This is unavoidable when changing persistence backends and is the motivation for moving to a properly migratable one going forward.

### DIFF
--- a/src/wallet/core.rs
+++ b/src/wallet/core.rs
@@ -4,7 +4,7 @@
 
 use super::*;
 
-const BDK_DB_NAME: &str = "bdk_db";
+const BDK_DB_NAME: &str = "bdk_db_sqlite";
 
 pub(crate) const NUM_KNOWN_SCHEMAS: usize = 4;
 
@@ -207,8 +207,8 @@ pub struct WalletInternals {
     pub(crate) _logger_guard: AsyncGuard,
     pub(crate) database: Arc<RgbLibDatabase>,
     pub(crate) wallet_dir: PathBuf,
-    pub(crate) bdk_wallet: PersistedWallet<Store<ChangeSet>>,
-    pub(crate) bdk_database: Store<ChangeSet>,
+    pub(crate) bdk_wallet: PersistedWallet<BdkStore>,
+    pub(crate) bdk_database: BdkStore,
     #[cfg(any(feature = "electrum", feature = "esplora"))]
     pub(crate) online_data: Option<OnlineData>,
 }
@@ -260,7 +260,7 @@ pub(crate) fn setup_bdk<P: AsRef<Path>>(
     desc_vanilla: String,
     watch_only: bool,
     bdk_network: BdkNetwork,
-) -> Result<(PersistedWallet<Store<ChangeSet>>, Store<ChangeSet>), Error> {
+) -> Result<(PersistedWallet<BdkStore>, BdkStore), Error> {
     let chain_net: ChainNet = wallet_data.bitcoin_network.into();
     let mut wallet_params = BdkWallet::load()
         .descriptor(KeychainKind::External, Some(desc_colored.clone()))
@@ -275,13 +275,16 @@ pub(crate) fn setup_bdk<P: AsRef<Path>>(
         BDK_DB_NAME.to_string()
     };
     let bdk_db_path = wallet_dir.as_ref().join(bdk_db_name);
-    let (mut bdk_database, _) =
-        Store::<ChangeSet>::load_or_create(BDK_DB_NAME.as_bytes(), bdk_db_path)?;
-    let bdk_wallet = match wallet_params.load_wallet(&mut bdk_database)? {
+    let bdk_db_url = format!("sqlite:{}", adjust_canonicalization(bdk_db_path));
+    let mut bdk_database = block_on(BdkStore::new(&bdk_db_url))?;
+    // schema migrations are applied automatically when the wallet is loaded
+    let bdk_wallet = match block_on(wallet_params.load_wallet_async(&mut bdk_database))? {
         Some(wallet) => wallet,
-        None => BdkWallet::create(desc_colored, desc_vanilla)
-            .network(bdk_network)
-            .create_wallet(&mut bdk_database)?,
+        None => block_on(
+            BdkWallet::create(desc_colored, desc_vanilla)
+                .network(bdk_network)
+                .create_wallet_async(&mut bdk_database),
+        )?,
     };
     Ok((bdk_wallet, bdk_database))
 }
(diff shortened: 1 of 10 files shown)


zoedberg:
Thank you for this PR, we really appreciate the detailed explanation of the dependency and migration tradeoffs.
FYI, I've added a warning about the BDK storage backwards compatibility issue to the 0.3.0 beta 6 release notes [here](https://github.com/RGB-Tools/rgb-lib/releases/tag/0.3.0-beta.6).
At the moment, we don't have the bandwidth to review this properly, but we will take a closer look in a couple of weeks. We'll get back to you as soon as we can. Thanks for your patience and contribution!

gofman8:
@dcorral could you please review this as well, thanks

src/wallet/core.rs
```diff
@@ -275,13 +275,16 @@ pub(crate) fn setup_bdk<P: AsRef<Path>>(
         BDK_DB_NAME.to_string()
     };
     let bdk_db_path = wallet_dir.as_ref().join(bdk_db_name);
-    let (mut bdk_database, _) =
-        Store::<ChangeSet>::load_or_create(BDK_DB_NAME.as_bytes(), bdk_db_path)?;
-    let bdk_wallet = match wallet_params.load_wallet(&mut bdk_database)? {
+    let bdk_db_url = format!("sqlite:{}", adjust_canonicalization(bdk_db_path));
+    let mut bdk_database = block_on(BdkStore::new(&bdk_db_url))?;
```
dcorral, line 279: Store::new uses the sqlx pool defaults, which allow up to 10 lazy connections. Since access to this store is serialized anyway, it could be worth building the pool explicitly and passing it in via Store::new_pool. Take the connection setup in setup_db just above as reference.

dcorral (changes requested):
Thanks for this PR. Moving off `bdk_file_store` makes a lot of sense, and the explanation of the dependency and migration trade-offs was really helpful while reviewing.

I left one small, note inline about the pool configuration.

zoedberg:
I have one concern about the long-term dependency coupling. The real issue isn't `sqlx` itself, but the shared native dependency on `libsqlite3-sys`. Because both `bdk_sqlite` (via `sqlx`) and `sea-orm` (also via `sqlx`) depend on that stack, Cargo requires them to resolve to a compatible version of `libsqlite3-sys`. As a result, our ability to upgrade either library is constrained by the other.

Within the current `sea-orm` 1.1.x / `sqlx` 0.8.x ecosystem, this isn't a problem because Cargo can resolve both to the same `sqlx`/`libsqlite3-sys` versions. However, the next major version bump could become a blocker. For example, if `sea-orm` 2.0.0 updates to a newer `sqlx`/`libsqlite3-sys` stack before `bdk_sqlite` does, we likely won't be able to upgrade to `sea-orm` 2.0 until `bdk_sqlite` catches up (or we maintain a fork/patch). The reverse is also true: if a future `bdk_sqlite` release moves to a newer `sqlx`/`libsqlite3-sys` stack before `sea-orm` does, we could be blocked from upgrading `bdk_sqlite` instead.

For this reason, I'm not convinced that what this PR proposes is the right long-term direction. I think I'd rather keep using `bdk_file_store` for now and handle future backwards-incompatible changes by discarding the existing BDK store and recreating it with a full scan sync. Ideally in the long-term I think we should use rgb-lib's DB for the BDK storage, that way when we'll add more DB management systems (e.g. PostreSQL) we will have rgb-lib and BDK data aligned to the same technology.

markuslevonyak:
Thanks for the detailed review; your point about the coupling is fair. Being stuck between `sea-orm` and `bdk_sqlite` on the next major bump is a real risk, and I agree it's a good reason not to take this PR as-is.

I do want to flag the problem that led me here, though, because I think it affects the "keep `bdk_file_store` and recreate on breakage" fallback.

After the BDK 2.0 to 3.0 upgrade, my BDK store was recreated and I got `UNIQUE constraint` errors from rgb-lib. The cause:

1. A `witness_receive` reveals address index N and stores its script in `pending_witness_script`, where `script` has a unique key.
2. That row is only removed once a matching UTXO is found, so an unpaid invoice keeps it indefinitely.
3. When the BDK store is recreated, the full scan only restores `last_revealed` up to roughly the last *used* index; revealed-but-unpaid addresses are never rediscovered.
4. The next `witness_receive` then re-reveals index N, derives the same script, and the insert fails on the unique constraint.

So for any wallet with open witness receives, discarding and rescanning the BDK store doesn't just lose cached data; it makes `witness_receive` fail until the collision is resolved. The revealed-address index isn't reconstructible from chain data, since those addresses have no on-chain history by definition.

That's why I don't think recreate-and-rescan works as a general recovery strategy, independent of which backend we choose.

Given that, your suggestion of using rgb-lib's own DB for BDK storage looks like the right direction to me, and it avoids the coupling problem entirely: with both `file_store` and `rusqlite` disabled, `bdk_wallet` has no native SQLite dependency, so `sea-orm` stays the only consumer of `libsqlite3-sys` and BDK data would follow rgb-lib to Postgres later. It's essentially a `WalletPersister` implementation over the existing connection.

Happy to close this PR and work on that instead if you agree.

zoedberg:
> After the BDK 2.0 to 3.0 upgrade, my BDK store was recreated and I got UNIQUE constraint errors from rgb-lib

Thanks to bringing this up. I thought the full scan was able to recover the BDK state completely.

> Happy to close this PR and work on that instead if you agree.

That would be great, thank you! Please let me know if you'll be able to add this before the next release, otherwise I'll work on a temporary solution to avoid similar issues in the future.

markuslevonyak:
> That would be great, thank you! Please let me know if you'll be able to add this before the next release, otherwise I'll work on a temporary solution to avoid similar issues in the future.

Great, I'll start working on this right away and will close this PR in favor of the new approach. I'm aiming to have it ready for the next release. Do you have a rough timeline for that?

zoedberg:
@markuslevonyak That's great, thank you! I think it will happen on the first or second week of August

PR #88, opened 2026-06-29
use advisory file lock for the RGB runtime lock

berserker-systems:
Replace the presence-based lockfile (created on acquire, removed on drop) with an OS advisory lock held for the runtime's lifetime. The kernel releases it automatically when the process dies, so a crash/kill no longer leaves a stale lock that must be deleted on start, which also restored proper mutual exclusion.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        toolchain: [ nightly, stable, 1.88.0 ]
+        toolchain: [ nightly, stable, 1.89.0 ]
         name: [linux, windows, macos]
         include:
           - name: linux
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ authors = [
     "Nicola Busanello <nicola.busanello@gmail.com>",
 ]
 edition = "2024"
-rust-version = "1.88.0"
+rust-version = "1.89.0"
 repository = "https://github.com/RGB-Tools/rgb-lib"
 homepage = "https://github.com/RGB-Tools/rgb-lib"
 license = "MIT"
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -697,8 +697,9 @@ impl ResolveWitness for DumbResolver {
 pub struct RgbRuntime {
     /// The RGB stock
     stock: Stock,
-    /// The wallet directory, where the lockfile for the runtime is to be held
-    wallet_dir: PathBuf,
+    /// File holding the advisory lock for the runtime's lifetime; the OS releases it when the file
+    /// is closed, including on process death, so no stale lock survives a crash.
+    _lock_file: fs::File,
 }
 
 impl RgbRuntime {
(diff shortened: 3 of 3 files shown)


zoedberg:
Thanks for this PR. We're currently a bit swamped, so we won't be able to review this right away. We plan to dive into it in a couple of weeks and will follow up then. Really appreciate your patience!

berserker-systems:
@zoedberg 
No worries and thank you for the notice.

As explanation: Android and iOS use SIGKILL sometimes to terminate a process. This will leave a stale lockfile.
At the moment, we remove it manually when the process starts. In my opinion this PR is the correct approach to handle the situation.

gofman8:
@dcorral could you please review this as well, thanks

dcorral (changes requested):
Thanks for this PR, this is the right approach. Kernel-released advisory locks solve the stale-lockfile-after-SIGKILL problem cleanly.

A few notes:
- The timeout error still says "unreleased lock file", with this change a timeout means live contention, not a stale file. Suggest rewording "timed out waiting for the RGB runtime lock" so users aren't tempted to delete the file, which would silently break exclusion for the next acquirer.
- Since the lockfile now persists forever, an older rgb-lib pointed at the same wallet dir will hang for the full timeout and fail until the file is manually deleted. Worth a line in the release notes @zoedberg, along with the MSRV bump to 1.89.
- The lockfile now always ends up in backups. Harmless since it's empty, but could be excluded alongside LOG_FILE.

zoedberg:
The problem you’re hitting is real and I understand why the current behavior is frustrating in production. However, I don’t think this is the right change for rgb-lib at this point.

**Why we’re declining**
Until we implement a journaling / error-recovery system (which is on our roadmap), an interrupted rgb-lib operation can leave the wallet in an inconsistent state. In that situation, the leftover `rgb_runtime.lock` file is intentional: it stops further use of the wallet and surfaces the problem instead of letting the app continue silently.

The advisory-lock approach in this PR would release the lock automatically when the process dies. That removes the stale-lock symptom, but it also makes post-interruption corruption much harder to notice, which we consider more dangerous than being blocked. In some cases, the wallet can be manually recovered to a healthy state; allowing the app to proceed unchecked reduces the chance that happens before more damage is done.

So for now, we prefer to keep the current fail-closed behavior rather than trade a UX inconvenience for a silent integrity risk.

**What we recommend instead**
rgb-lib cannot prevent the host OS from killing the app or an rgb-lib subprocess (e.g. via SIGKILL). If the process is force-terminated mid-operation, that’s outside what the library can intercept or recover from today.

While a proper recovery system is not yet in place, wallet/app developers should handle this at the application level. For example:
- run rgb-lib in a long-lived service process (background service, worker, daemon, etc.) so it outlives the UI and can finish in-flight work even when the foreground app is closed or backgrounded
- use lifecycle hooks and signal handling (SIGTERM where available, iOS/Android lifecycle callbacks, graceful shutdown paths) to stop accepting new work and let the current operation complete before the process is actually killed

If your app hits a stale lock after a force-close, the safe path today is to treat it as a wallet that may need recovery, not to automatically clear the lock and continue. We can document this guidance more explicitly if that would help downstream integrators.

Sorry this isn’t the fix you were hoping for and thanks again for raising the issue. We’ll revisit locking behavior once that infrastructure is in place.

berserker-systems:
@zoedberg

Thank you very much for the detailed explanation. Sadly... you have a very strong argument there.
Do you have plans to add journaling / error-recovery soon?

zoedberg:
@berserker-systems I cannot quantify "soon", but yes, it's on our roadmap, I really hope we can address this right after the other major changes we are working on.

PR #90, opened 2026-07-09
fix: recover witness UTXOs orphaned by a lost consignment

berserker-systems:
A witness invoice can be paid on-chain without its consignment ever being delivered: the proxy can drop a successfully posted consignment (e.g. it restarts and loses its file store), and the sender may then pay the same invoice again with a new TX whose consignment does arrive. This only happens in donation mode, where the sender broadcasts right after posting the consignment; otherwise it waits for the recipient's ACK (which requires the recipient to already hold the consignment) before broadcasting, so a lost consignment cannot leave a confirmed orphan. The first payment's UTXO is quarantined with pending_witness set, but no transfer ever references its TXID, so the flag is never cleared: the UTXO stays unspendable forever and its allocation, received as part of the second consignment's history, lives only in the RGB runtime and never makes it into the DB (understated balance, sends failing with InsufficientAllocationSlots).

On refresh, look for pending witness TXOs with no transfer referencing their TXID. When the runtime already holds allocations for one there is nothing left to wait for: save them to the DB as a settled transfer and clear the flag, so both the sats and the assets become usable again. TXOs the runtime knows nothing about are left as-is, since spending them would burn whatever a still-pending consignment may deliver.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -248,7 +248,7 @@ use crate::utils::INDEXER_BATCH_SIZE;
 use crate::utils::INDEXER_PARALLEL_REQUESTS;
 #[cfg(any(feature = "electrum", feature = "esplora"))]
 #[cfg(test)]
-use crate::wallet::test::{mock_input_unspents, mock_vout};
+use crate::wallet::test::{mock_consignment_recipient_id, mock_input_unspents, mock_vout};
 #[cfg(any(feature = "electrum", feature = "esplora"))]
 use crate::{
     api::{
--- a/src/wallet/online.rs
+++ b/src/wallet/online.rs
@@ -1760,6 +1760,113 @@ pub trait WalletOnline: WalletOffline {
         Ok(refresh_result)
     }
 
+    /// Reconcile orphaned pending witness TXOs with the RGB runtime.
+    ///
+    /// If no consignment ever references the TXID of a TXO in the pending witness state (e.g.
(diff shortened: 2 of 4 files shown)


zoedberg:
Thank you for this PR and for the clear write-up of the orphan witness UTXO scenario. The problem is real and the test covers the failure path well.

We're about to release a large set of changes, including a refactor of how the proxy integration works. We'd prefer not to merge this now, since the surrounding code is likely to change significantly. Could you please rebase this PR once those changes are public? We can then reassess whether this reconciliation logic still works.

On that point: the transport proxy is chosen by the receiver, and in practice it should ideally be infrastructure the receiver controls. Relying on third-party proxies is possible, but should be done carefully: users should understand and trust each proxy's data retention and availability guarantees. With a receiver-owned proxy, the failure mode this PR addresses (a consignment posted successfully but later unavailable) becomes much less likely, and we may find that explicit orphan reconciliation is unnecessary altogether.

That said, donation mode remains inherently risky regardless of proxy setup. It broadcasts before the recipient has accepted the consignment, which creates several ways things can go wrong, for example the recipient may consider the consignment invalid and refuse it. Wallets using donation mode should treat it as an explicit, operator-controlled option, not a default path, and be aware of these risks.

Thanks again for the contribution. We'll revisit this once our upcoming changes are out.

berserker-systems:
@zoedberg Thank you. Please ping or e-mail me, when we can continue here.
Quick note: this also happens, when only btc gets send to an rgb-invoice address.

PR #92, opened 2026-07-27
Wallet open fails permanently on a corrupt BDK file store

dcorral:
## Problem

BDK's changeset store is an append-only `bincode` log. If a `persist` is interrupted (crash, power loss, kill during write), the file can be left with an undecodable trailing record. From that point on, `Store::load_or_create` fails on every wallet open — the wallet is bricked until the user manually deletes or repairs the store file, even though all the data written before the torn record is intact and BDK itself can still return it.

### DIFF
--- a/src/wallet/test/new.rs
+++ b/src/wallet/test/new.rs
@@ -623,3 +623,62 @@ fn supported_schemas() {
         assert_matches!(e, Error::CannotUseIfaOnMainnet);
     }
 }
+
+/// A torn trailing record in the BDK file store (e.g. a `persist` interrupted
+/// by a crash or power loss) must not brick the wallet: the store data written
+/// before the bad record is intact and BDK returns it alongside the corruption
+/// error, so wallet open should recover the store and succeed instead of
+/// failing on every attempt until the file is repaired by hand.
+#[test]
+#[parallel]
+fn recovers_from_torn_bdk_store_tail() {
+    use std::io::Write;
+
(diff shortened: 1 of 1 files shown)


dcorral:
Proposed fix is wrap the Store::load_or_create call in setup_bdk with a recovery path:

- corruption error with recoverable changeset -> rebuild store from the changeset BDK returns, atomic-swap into place, keep original as <name>.corrupt[.N], log warning, open normally with all pre-corruption state intact
- corruption with nothing recoverable -> fail as today, file left untouched.

This was already implemented in the UTEXO fork (https://github.com/UTEXO-Protocol/rgb-lib/pull/44)

zoedberg:
The proposed fix is imprecise and incomplete. Recovering a BDK partial state doesn't guarantee that it will contain all that rgb-lib expects to be there.

A better fix would be to just recreate the BDK storage when having issues loading it, with some quirks because in order for the wallet to behave correctly after recreating the BDK storage you also need to:
- save in rgb-lib the last revealed index for both keychains (without those a BDK sync would not recover the BDK state completely)
- save in rgb-lib's DB that BDK storage has been deleted and no full scan sync has been performed yet (needed to disable APIs that cannot function properly until the sync is performed)
 
In https://github.com/RGB-Tools/rgb-lib/pull/87 @markuslevonyak has proposed to work on a better solution, which is stop using `bdk_file_store`, that is not production ready, and move BDK storage in rgb-lib's DB. So we're waiting for that to land instead of proceeding with a temporary solution like the one you proposed.

PR #93, opened 2026-07-27
Wallet sync leaves rgb-lib DB and BDK store diverged on colored TXOs

dcorral:
rgb-lib keeps its own `Txo` table alongside BDK's persisted wallet store. The two can diverge on whether a colored UTXO is still spendable: rgb-lib may hold a row as `exists && !spent && !pending_witness` while the same outpoint is no longer in BDK's `list_unspent`. This can happen after restoring the wallet directory from a snapshot taken at a different chain state, an out-of-band spend of a wallet outpoint, or a non-graceful shutdown that persists one store but not the other.

Once diverged, nothing heals the state, the only code path that ever marks a `Txo` row spent is inside `broadcast_tx`, i.e. only for inputs of a tx this wallet instance broadcasts itself. The wallet keeps operating with a DB that claims spendable colored allocations backed by outpoints BDK no longer considers unspent, so UTXO selection, balance reporting and sends work from inconsistent data indefinitely.

### DIFF
--- a/src/wallet/test/sync.rs
+++ b/src/wallet/test/sync.rs
@@ -40,3 +40,253 @@ fn fail() {
     party.online = good_online;
     assert!(matches!(result, Err(Error::CannotChangeOnline)));
 }
+
+/// Sync reconciles rgb-lib's Txo table downward: a row that rgb-lib holds as
+/// `exists=true && spent=false && pending_witness=false` but whose outpoint is
+/// not in BDK's `list_unspent` after sync gets marked spent.
+///
+/// Simulates the divergence where an out-of-band event (snapshot restore,
+/// manual edit, second process) left rgb-lib's DB claiming a UTXO that BDK no
+/// longer indexes.
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn sync_marks_orphan_txo_spent() {
+    initialize();
+
+    let mut wallet = get_test_wallet(true, None);
+    let online = wallet.go_online(test_go_online_options(None)).unwrap();
+
+    // Insert an orphan Txo row: rgb-lib's DB claims this outpoint is on-chain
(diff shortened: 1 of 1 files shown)


dcorral:
Proposed fix is a reconcile pass on the `Wallet::sync` path (colored keychain):

- any `Txo` row held as `exists && !spent && !pending_witness` whose outpoint is not in BDK's `list_unspent` after sync -> marked spent
- one-directional by design: rows are only ever marked spent, never revived to unspent, so a UTXO already observed as consumed cannot come back
- `pending_witness` and `exists=false` rows are skipped (not yet on-chain from the wallet's perspective, not "missing")
- deliberately not run inside `sync_bdk_and_db_txos`, so the `go_online` consistency check still sees the raw divergence and can refuse to come online in the cross-device-wallet case

This was already implemented in the UTEXO fork (https://github.com/UTEXO-Protocol/rgb-lib/pull/42)

zoedberg:
> This can happen after restoring the wallet directory from a snapshot taken at a different chain state

Restore of a partial snapshot is not something that's implemented in master. You proposed that in https://github.com/RGB-Tools/rgb-lib/pull/92#issuecomment-5095593087 but as explained there the fix is incomplete and there's a better solution that is being implemented. Without that incomplete fix the condition described here should never happen. 

> an out-of-band spend of a wallet outpoint

This would break rgb-lib, it's documented in the README that using rgb-lib's mnemonic with other wallets or on multiple devices should never be done. Unfortunately we cannot prevent a user from getting the mnemonic and using it outside, but we cannot even protect against every wrong user behavior, so I don't think rgb-lib should try to solve issues caused by an incorrect usage.

> a non-graceful shutdown that persists one store but not the other

This will also be addressed by putting the BDK storage in rgb-lib's DB (thanks to DB transitions all data will be either saved or dropped, even in case of a panic).

PR #97, opened 2026-08-06
list_transfers to take an AssetFilter and an optional txid

dcorral:
Breaking change to list_transfers, replacing the asset_id: Option<String> parameter (where None meant "transfers not connected to a specific asset") with two explicit parameters:

- filter: AssetFilter. Any (all transfers), NoAsset (transfers not connected to a specific asset, the old None behavior), or Id(asset_id) (the old Some behavior, still validated with AssetNotFound)
- txid: Option<String>. When set, restricts the result to the transfers committed by that on-chain transaction; combined with an asset filter it acts as an intersection; an unknown txid returns an empty list

This makes it possible to look up the transfers behind a given on-chain transaction without listing per asset and filtering client-side.

### DIFF
--- a/bindings/uniffi/src/lib.rs
+++ b/bindings/uniffi/src/lib.rs
@@ -12,10 +12,10 @@ use rgb_lib::{
     keys::{Keys, WitnessVersion},
     utils::BitcoinNetwork,
     wallet::{
-        Address as RgbLibAddress, AssetCFA, AssetIFA, AssetNIA, AssetUDA, Assets,
-        AssignmentsCollection, Balance, BlockTime, BtcBalance, BurnBeginResult, BurnDetails,
-        Cosigner as CosignerData, DatabaseType, EmbeddedMedia, HubInfo, InflateBeginResult,
-        InflateDetails, InitOperationResult, Invoice as RgbLibInvoice,
+        Address as RgbLibAddress, AssetCFA, AssetFilter as RgbLibAssetFilter, AssetIFA, AssetNIA,
+        AssetUDA, Assets, AssignmentsCollection, Balance, BlockTime, BtcBalance, BurnBeginResult,
+        BurnDetails, Cosigner as CosignerData, DatabaseType, EmbeddedMedia, HubInfo,
+        InflateBeginResult, InflateDetails, InitOperationResult, Invoice as RgbLibInvoice,
         InvoiceData as RgbLibInvoiceData, Media, Metadata, MultisigKeys, MultisigOnlineOptions,
         MultisigVotingStatus as RgbLibMultisigVotingStatus, MultisigWallet as RgbLibMultisigWallet,
         Online, OnlineOptions, Operation as RgbLibOperation, OperationInfo as RgbLibOperationInfo,
@@ -67,6 +67,21 @@ impl From<SyncOptions> for RgbLibSyncOptions {
         Self {
             keychain: orig.keychain.into(),
             strategy: orig.strategy,
+        }
+    }
+}
+
+pub enum AssetFilter {
+    AnyOrNone,
+    None,
+    Id { asset_id: String },
+}
+impl From<AssetFilter> for RgbLibAssetFilter {
+    fn from(orig: AssetFilter) -> Self {
+        match orig {
+            AssetFilter::AnyOrNone => RgbLibAssetFilter::AnyOrNone,
+            AssetFilter::None => RgbLibAssetFilter::None,
+            AssetFilter::Id { asset_id } => RgbLibAssetFilter::Id(asset_id),
         }
     }
 }
@@ -1296,10 +1311,14 @@ impl Wallet {
         self._get_wallet().list_transactions(online, skip_sync)
     }
 
-    fn list_transfers(&self, asset_id: Option<String>) -> Result<Vec<Transfer>, RgbLibError> {
+    fn list_transfers(
+        &self,
+        asset_filter: AssetFilter,
+        txid: Option<String>,
+    ) -> Result<Vec<Transfer>, RgbLibError> {
         Ok(self
             ._get_wallet()
-            .list_transfers(asset_id)?
+            .list_transfers(asset_filter.into(), txid)?
             .into_iter()
             .map(|t| t.into())
             .collect())
@@ -1741,10 +1760,14 @@ impl MultisigWallet {
         self._get_wallet().list_transactions(online, skip_sync)
     }
 
-    fn list_transfers(&self, asset_id: Option<String>) -> Result<Vec<Transfer>, RgbLibError> {
+    fn list_transfers(
+        &self,
+        asset_filter: AssetFilter,
+        txid: Option<String>,
+    ) -> Result<Vec<Transfer>, RgbLibError> {
         Ok(self
             ._get_wallet()
-            .list_transfers(asset_id)?
+            .list_transfers(asset_filter.into(), txid)?
             .into_iter()
             .map(|t| t.into())
             .collect())
--- a/src/wallet/objects.rs
+++ b/src/wallet/objects.rs
@@ -181,6 +181,17 @@ pub struct BlockTime {
 // Assets, tokens & media
 // ────────────────────────────────────────────────────────────
 
+/// An asset filter.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub enum AssetFilter {
+    /// Match any or no asset
+    AnyOrNone,
+    /// Match no asset
+    None,
+    /// Match the asset with the given ID
+    Id(String),
+}
+
 /// An asset media file.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize, Serialize)]
 #[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
--- a/src/wallet/offline.rs
+++ b/src/wallet/offline.rs
@@ -2087,13 +2087,31 @@ pub trait WalletOffline: WalletBackup {
     fn list_transfers_impl(
         &self,
         txn: &DbTxn,
-        asset_id: Option<String>,
+        asset_filter: AssetFilter,
+        txid: Option<String>,
     ) -> Result<Vec<Transfer>, Error> {
         let db_data = txn.get_db_data(false)?;
+        let batch_transfer_idxs: Option<HashSet<i32>> = txid.map(|txid| {
+            db_data
+                .batch_transfers
+                .iter()
+                .filter(|b| b.txid.as_deref() == Some(txid.as_str()))
+                .map(|b| b.idx)
+                .collect()
+        });
         let asset_transfer_ids: Vec<i32> = db_data
             .asset_transfers
             .iter()
-            .filter(|t| t.asset_id == asset_id)
+            .filter(|t| match &asset_filter {
+                AssetFilter::AnyOrNone => true,
+                AssetFilter::None => t.asset_id.is_none(),
+                AssetFilter::Id(asset_id) => t.asset_id.as_ref() == Some(asset_id),
+            })
+            .filter(|t| {
+                batch_transfer_idxs
+                    .as_ref()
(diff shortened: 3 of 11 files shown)


src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
     pub(crate) psbt_path: Option<String>,
 }
 
+/// An asset filter for listing transfers.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
+pub enum AssetFilter {
```
zoedberg: please move this at the beginning of the " Assets, tokens & media" section

bindings/uniffi/src/lib.rs
```diff
@@ -98,6 +98,20 @@ impl From<Assignment> for RgbLibAssignment {
         }
     }
 }
+pub enum AssetFilter {
```
zoedberg: this is not the right place for this, you are inside a "// temporary solution needed because the Enum attribute doesn't support the Remote one" section related to the `Assignment`

src/database/mod.rs
```diff
@@ -518,14 +518,13 @@ impl DbTxn {
...
 
-    #[cfg(any(feature = "electrum", feature = "esplora"))]
     pub(crate) fn get_batch_transfer_by_txid(
         &self,
         txid: &str,
     ) -> Result<Option<DbBatchTransfer>, Error> {
         Ok(block_on(
             BatchTransfer::find()
-                .filter(batch_transfer::Column::Txid.eq(txid))
+                .filter(crate::database::entities::batch_transfer::Column::Txid.eq(txid))
```
zoedberg:
```suggestion
                .filter(batch_transfer::Column::Txid.eq(txid))
```

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
     pub(crate) psbt_path: Option<String>,
 }
 
+/// An asset filter for listing transfers.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
```
zoedberg: enums should not use camelCase (I think there are 3 similar mistakes in multisig.rs if you could drop them too)

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
     pub(crate) psbt_path: Option<String>,
 }
 
+/// An asset filter for listing transfers.
```
```suggestion
/// An asset filter.
```

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
...
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
+pub enum AssetFilter {
+    /// Match transfers of any asset, including those not connected to one
```
```suggestion
    /// Match any or no asset
```

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
...
+    Any,
+    /// Match only transfers not connected to a specific asset
```
```suggestion
    /// Match no asset
```

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
...
+    NoAsset,
+    /// Match only transfers of the asset with the given ID
```
```suggestion
    /// Match the asset with the given ID
```

src/wallet/offline.rs
```diff
@@ -2087,13 +2087,26 @@ pub trait WalletOffline: WalletBackup {
     fn list_transfers_impl(
         &self,
         txn: &DbTxn,
-        asset_id: Option<String>,
+        filter: AssetFilter,
```
zoedberg:
```suggestion
        asset_filter: AssetFilter,
```
please do the same rename in the bindings

src/wallet/offline.rs
```diff
@@ -2719,18 +2732,23 @@ pub trait RgbWalletOpsOffline: WalletOffline + WalletBackup {
 
     /// List the RGB [`Transfer`]s known to the wallet.
     ///
-    /// When an `asset_id` is not provided, return transfers that are not connected to a specific
-    /// asset.
-    fn list_transfers(&self, asset_id: Option<String>) -> Result<Vec<Transfer>, Error> {
+    /// `filter` selects transfers by asset. When a `txid` is provided, restrict the result to the
```
zoedberg:
```suggestion
    /// `asset_filter` selects transfers by asset. When a `txid` is provided, restrict the result to the
```

src/wallet/offline.rs
```diff
@@ -2719,18 +2732,23 @@ pub trait RgbWalletOpsOffline: WalletOffline + WalletBackup {
...
+    /// transfers committed by the on-chain transaction with that ID; an unknown `txid` yields an
+    /// empty list.
+    fn list_transfers(
+        &self,
+        filter: AssetFilter,
```
zoedberg:
```suggestion
        asset_filter: AssetFilter,
```

src/wallet/test/utils/api.rs
```diff
@@ -502,7 +502,7 @@ pub(crate) trait OfflineSigParty {
     #[cfg(feature = "electrum")]
     fn get_pending_blind_transfers(&self) -> Vec<Transfer> {
         self.wlt()
-            .list_transfers(None)
+            .list_transfers(AssetFilter::NoAsset, None)
```
zoedberg:
```suggestion
            .list_transfers(AssetFilter::AnyOrNone, None)
```

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
     pub(crate) psbt_path: Option<String>,
 }
 
+/// An asset filter for listing transfers.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[cfg_attr(feature = "camel_case", serde(rename_all = "camelCase"))]
+pub enum AssetFilter {
+    /// Match transfers of any asset, including those not connected to one
+    Any,
```
```suggestion
    AnyOrNone,
```

src/wallet/objects.rs
```diff
@@ -1233,6 +1233,18 @@ pub struct TransferData {
...
+    /// Match only transfers not connected to a specific asset
+    NoAsset,
```
```suggestion
    None,
```

src/wallet/test/list_transfers.rs
```diff
@@ -166,6 +166,105 @@ fn success() {
...
+        .collect();
+    assert_eq!(vec![nia_by_txid[0].idx], expected);
+
+    // Any + no txid: the whole history (2 issuances + 2 sends)
+    let all = party.list_transfers_filtered(AssetFilter::Any, None);
+    assert_eq!(all.len(), 4);
+
+    // NoAsset: receiver's pending blind receives not yet tied to an asset
+    let pending = rcv_party.list_transfers_filtered(AssetFilter::NoAsset, None);
+    assert_eq!(pending.len(), 2);
```
zoedberg: after this do a refresh for `rcv_party` of just one of the 2 transfers, then check again what the `list_transfers_filtered(AssetFilter::NoAsset, None)` call returns, it should have len 1

src/wallet/test/list_transfers.rs
```diff
@@ -174,4 +273,11 @@ fn fail() {
     // asset not found
     let result = party.list_transfers_result(Some("rgb1inexistent"));
     assert!(matches!(result, Err(Error::AssetNotFound { asset_id: _ })));
+
+    // asset not found also when a txid is given
+    let result = party.list_transfers_filtered_result(
+        AssetFilter::Id(s!("rgb1inexistent")),
+        Some("0000000000000000000000000000000000000000000000000000000000000000"),
```
zoedberg: use the `FAKE_TXID` variable here

zoedberg (changes requested):
Thanks! Please address the requested changes

src/wallet/test/list_transfers.rs
dcorral: I tried but didn't find a proper way to refresh just one of the 2 transfers since a RefreshFilter can't distinguish them and refreshing with an asset_id doesn't work at this point since the receiver doesn't know the assets yet. 

A plain refresh picks up both consignments in one call, so for now I added the check right after it, asserting the no-asset list is empty. If there's a way to refresh a single transfer that I'm missing, happy to change it to the len 1 check you suggested.

zoedberg: You're right, I forgot there's no way to refresh a single transfer, sorry for that.

The reason I asked for the len 1 check is that the current assertions don't really pin down the filter. At the len == 2 check the receiver's whole transfer set is those 2 no-asset receives, so `None` and `AnyOrNone` are indistinguishable there and a filter ignoring its argument would pass.

We can get the stronger check without refreshing a single transfer: add a third `blind_receive` on `rcv_party` that nothing is sent to, before the refresh. After the refresh the receiver holds 3 transfers, 2 tied to their assets and 1 not, so:
```rust
let receive_extra = rcv_party.blind_receive();   // stays asset-less
// ... refresh ...
assert_eq!(rcv_party.list_transfers_filtered(AssetFilter::AnyOrNone, None).len(), 3);
let pending = rcv_party.list_transfers_filtered(AssetFilter::None, None);
assert_eq!(pending.len(), 1);
assert_eq!(pending[0].recipient_id, Some(receive_extra.recipient_id));
```
That way a single call has to partition a mixed set, and the identity check makes sure it's the right transfer and not a count that happens to match.

src/wallet/offline.rs
```diff
@@ -2087,13 +2087,26 @@ pub trait WalletOffline: WalletBackup {
     fn list_transfers_impl(
         &self,
         txn: &DbTxn,
-        asset_id: Option<String>,
+        asset_filter: AssetFilter,
+        txid: Option<String>,
     ) -> Result<Vec<Transfer>, Error> {
+        let batch_transfer_idx = match txid {
+            Some(txid) => match txn.get_batch_transfer_by_txid(&txid)? {
```
zoedberg: I just noticed that `get_batch_transfer_by_txid` calls `.one()`, but `txid` isn't unique across incoming batch transfers, so we are silently dropping results here.

The new test doesn't catch it because the txid checks all run on `party`, where the batch tx is a single row. Please add a receiver-side assertion to the test too: after the refresh, `rcv_party.list_transfers_filtered(AssetFilter::AnyOrNone, Some(&txid))` should have len 2.

I will then prove another already-existing bug and do the fix on master and ask you to rebase this PR on top of it

zoedberg (changes requested):

src/wallet/offline.rs
zoedberg: @dcorral thank you, I see you modified the test which now as expected fails. I merged https://github.com/RGB-Tools/rgb-lib/pull/98 in the meantime, so you can now rebase this PR on top of the updated master tip, it should make the test pass

src/wallet/offline.rs
```diff
@@ -2087,13 +2087,37 @@ pub trait WalletOffline: WalletBackup {
...
+                    .map(|b| b.idx)
+                    .collect();
+                if idxs.is_empty() {
+                    return Ok(vec![]);
+                }
+                Some(idxs)
+            }
+            None => None,
+        };
         let db_data = txn.get_db_data(false)?;
```
zoedberg:
```suggestion
        let db_data = txn.get_db_data(false)?;
        let batch_transfer_idxs: Option<HashSet<i32>> = txid.map(|txid| {
            db_data
                .batch_transfers
                .iter()
                .filter(|b| b.txid.as_deref() == Some(txid.as_str()))
                .map(|b| b.idx)
                .collect()
        });
```

zoedberg (changes requested):
let's use `db_data`, to avoid the extra query

note that this change will make you revert the ones you did to `src/database/mod.rs`

zoedberg (approved):
LGTM, thank you!

PR #99, opened 2026-08-11
c-ffi: add missing begin/end exports and string free

dcorral:


bindings/c-ffi/src/lib.rs
```diff
@@ -515,6 +591,11 @@ pub extern "C" fn rgblib_sign_psbt(
     sign_psbt(wallet, unsigned_psbt).into()
 }
 
+#[unsafe(no_mangle)]
+pub extern "C" fn rgblib_string_free(ptr: *mut c_char) {
```
zoedberg: this is needed only for C and C++ bindings since Swig already frees strings. to prove that this is necessary you should add to the `example.c` file a method similar to `checkMemoryLeak` in the node.js example, in a separate commit, before adding the method to free strings

zoedberg (changes requested):
thanks, please see the requested changes

bindings/c-ffi/src/lib.rs
dcorral: done, the example now has a checkMemoryLeak method (added in a separate commit before the free function). Also fixed the example's list_transfers call, which no longer compiled after the asset-filter change.

zoedberg: @dcorral did you check if memory leak issues really existed and if the free string utility solved them?

dcorral: actually yes I measured over 10_000 rgblib_generate_keys calls and RSS grows ~0.5kB/call without freeing but stays flat with rgblib_string_free.

dcorral: I used a standalone variant of the example's checkMemoryLeak that samples VmRSS from /proc/self/status every 1000 iterations, with a flag to toggle freeing:

<details>
<summary><code>leak_harness.c</code></summary>

#include "rgblib.h"
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static long vmrss_kb(void) {
    FILE *f = fopen("/proc/self/status", "r");
    char line[256];
    long kb = -1;
    while (fgets(line, sizeof(line), f)) {
        if (strncmp(line, "VmRSS:", 6) == 0) {
            sscanf(line + 6, "%ld", &kb);
            break;
        }
    }
    fclose(f);
    return kb;
}

int main(int argc, char *argv[]) {
    int do_free = argc > 1 && strcmp(argv[1], "free") == 0;
    int iters = argc > 2 ? atoi(argv[2]) : 5000;
    printf("mode=%s iters=%d\n", do_free ? "free" : "leak", iters);
    printf("start RSS: %ld kB\n", vmrss_kb());
    for (int i = 1; i <= iters; i++) {
        CResultString r = rgblib_generate_keys("Regtest", "Taproot");
        if (r.result == Err) {
            printf("ERR: %s\n", r.inner);
            return 1;
        }
        if (i == 1)
            printf("returned string length: %zu bytes\n", strlen(r.inner));
        if (do_free)
            rgblib_string_free(r.inner);
        if (i % 1000 == 0)
            printf("iter %d RSS: %ld kB\n", i, vmrss_kb());
    }
    printf("end RSS: %ld kB\n", vmrss_kb());
    return 0;
}
</details>

Without freeing RSS grows linearly, +548 kB per 1000 calls, matching the returned string size:

mode=leak iters=10000
start RSS: 4880 kB
returned string length: 542 bytes
iter 1000 RSS: 8248 kB
iter 2000 RSS: 8800 kB
iter 3000 RSS: 9348 kB
iter 4000 RSS: 9896 kB
...
iter 10000 RSS: 13184 kB
end RSS: 13184 kB

With rgblib_string_free, RSS is flat after warm-up:

mode=free iters=10000
start RSS: 4760 kB
returned string length: 541 bytes
iter 1000 RSS: 7604 kB
iter 2000 RSS: 7608 kB
iter 3000 RSS: 7608 kB
...
iter 10000 RSS: 7128 kB
end RSS: 7128 kB

Verified with LeakSanitizer. Without freeing it reports precisely one leaked string per call:

Direct leak of 546901 byte(s) in 1000 object(s) allocated from:
SUMMARY: AddressSanitizer: 546901 byte(s) leaked in 1000 allocation(s).

With rgblib_string_free: no leaks reported. I also ran the checkMemoryLeak from example.c itself under AddressSanitizer, clean, zero leaks.

zoedberg: Thanks for checking!

zoedberg: since this is not an API of rgb-lib, I would call it `free_string` and place it before `free_wallet`. also, no need for the logic to live in `utils.rs`

zoedberg (changes requested):

zoedberg (approved):
LGTM, thanks!

PR #103, opened 2026-08-18
Add RGB contract consignment import and export

### LINKED ISSUE #56: Add method to export contract consignment
Introduce a method to export contract consignments in the standard format
```
-----BEGIN RGB CONSIGNMENT-----
Id: 
Version: 
Type:
Contract:
Schema:
Interface: 
Check-SHA256: 

...
-----END RGB CONSIGNMENT-----
```

Jainakin:
Closes #56

## Problem

Asset registries need a standard RGB contract consignment that can be verified and imported independently of an asset transfer. An asset ID alone cannot reconstruct or validate the contract, while a transfer consignment carries ownership history that is not appropriate for contract preload.

## Changes

- re-export `RgbContract` and add `Wallet::export_asset_contract`
- add one `Wallet::import_asset_contract(contract, media_file_paths)` API returning the imported asset `Metadata`
- validate the contract schema, wallet network, and contract consignment before mutation
- support NIA, CFA, IFA, and UDA contract metadata and declared media
- make repeated imports idempotent when the asset already exists in the wallet database
- verify that exported contracts contain no transfer bundles or terminals

Both APIs follow the existing rust-only feature model and are available with the `electrum` or `esplora` feature. The exported value uses rgb-ops' standard ASCII-armored `RGB CONSIGNMENT` `Display`/`FromStr` representation.

## Implementation notes

- contract-only import registers public contract identity and metadata; it creates no allocation, transfer, or wallet ownership, so the imported balance is zero
- rgb-ops rejects bundles and terminals in a contract consignment before witness resolution; the import uses the existing `DumbResolver` required by the validation API
- RGB Stock remains the persistence authority: `Stock::import_contract` uses its existing autosave transaction behavior
- media ingestion reuses the existing `file_details`, `copy_media_file`, and `cleanup_media_files` helpers; supplied files are matched to contract attachments by committed digest, while declared media metadata remains authoritative
- database metadata and backup state use the existing `save_new_asset_internal`, `get_asset_metadata_impl`, and `update_backup_info` paths in one database transaction
- the RGB runtime is released before metadata is read so UDA metadata can reopen the persisted runtime without contending on the wallet lock
- the existing `save_new_asset` transfer-consignment API and behavior are unchanged

## Review scope correction

The review-requested scope reduction is complete. The patch was rebuilt on current `master` and reduced from 7 files / 1,033 additions to 5 files / 428 additions. It removes:

- custom `RgbRuntime` persistence flags and drop behavior
- unrelated read-path and offline-build changes
- `ImportAssetContractResult` and the media wrapper type
- the repair/state-detection protocol
- duplicated media persistence and fsync code
- standalone repair, lock-contention, and duplicated transfer-cycle tests

## Validation

Local validation on the current head (`2741520`) is green:

- formatting, including formatted doc comments
- strict Clippy with all features/all targets, no default features/all targets, Electrum-only/all targets, and Esplora-only/all targets
- no-default-feature suite: 49 passed
- documentation tests: 2 passed
- contract-focused Docker tests: 4 passed
- modified end-to-end `send::success`: passed

The full GitHub Actions matrix was green on the preceding head (`210a15f`). Fresh workflow runs were created for `2741520`, but GitHub currently marks them `action_required`; a maintainer must approve the fork workflows before the authoritative Linux matrix can run.

The local Docker harness's full-service startup is not portable to this Apple Silicon host because its optional Esplora fixture is amd64-only. The focused integration tests were therefore run against the repository's isolated bitcoind, primary Electrum, RGB proxy, and modified-proxy fixtures. No test-harness changes are part of this PR.

## Coverage

Tests cover canonical armor round-trip, all four supported schemas, declared media, missing/unexpected/duplicate media, wrong-network rejection, malformed and unknown asset IDs, zero-balance semantics, idempotent re-import, export after a settled transfer, and rejection of history-bearing contract input.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,7 +106,8 @@ pub use bdk_wallet::bitcoin;
 pub use rgbstd::{
     ContractId, Txid as RgbTxid,
     containers::{
-        ConsignmentExt, Fascia, FileContent, PubWitness, Transfer as RgbTransfer, WitnessBundle,
+        ConsignmentExt, Contract as RgbContract, Fascia, FileContent, PubWitness,
+        Transfer as RgbTransfer, WitnessBundle,
     },
     persistence::UpdateRes,
     schema::SchemaId,
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -719,6 +719,16 @@ pub struct RgbRuntime {
 
 impl RgbRuntime {
     #[cfg(any(feature = "electrum", feature = "esplora"))]
+    pub(crate) fn export_contract(
+        &self,
+        contract_id: ContractId,
+    ) -> Result<RgbContract, InternalError> {
+        self.stock
+            .export_contract(contract_id)
+            .map_err(InternalError::from)
+    }
+
+    #[cfg(any(feature = "electrum", feature = "esplora"))]
     pub(crate) fn accept_transfer<R: ResolveWitness>(
         &mut self,
         contract: ValidTransfer,
--- a/src/wallet/rust_only.rs
+++ b/src/wallet/rust_only.rs
@@ -24,6 +24,23 @@ pub struct ColoringInfo {
     pub nonce: Option<u64>,
 }
 
+#[cfg(any(feature = "electrum", feature = "esplora"))]
+struct ContractImportResolver;
+
+#[cfg(any(feature = "electrum", feature = "esplora"))]
+impl ResolveWitness for ContractImportResolver {
+    fn resolve_witness(&self, witness_id: RgbTxid) -> Result<WitnessStatus, WitnessResolverError> {
+        Err(WitnessResolverError::ResolverIssue(
+            Some(witness_id),
+            s!("contract consignments cannot contain witness history"),
+        ))
+    }
+
+    fn check_chain_net(&self, _: ChainNet) -> Result<(), WitnessResolverError> {
+        Ok(())
+    }
+}
+
 /// Map of contract ID and list of its beneficiaries
 pub type AssetBeneficiariesMap = BTreeMap<ContractId, Vec<BuilderSeal<GraphSeal>>>;
 
@@ -79,6 +96,208 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
 
 /// Rust-only APIs of the wallet.
 impl Wallet {
+    /// Export an RGB contract known to this wallet.
+    ///
+    /// The returned value is a standard RGB contract consignment. It contains the public contract
+    /// definition, not transfer history, wallet allocations or proof that this wallet owns units
+    /// of the asset.
+    ///
+    /// <div class="warning">This method is meant for special usage and is normally not needed,
+    /// use it only if you know what you're doing</div>
+    #[cfg(any(feature = "electrum", feature = "esplora"))]
+    pub fn export_asset_contract(&self, asset_id: String) -> Result<RgbContract, Error> {
+        info!(self.logger(), "Exporting asset contract '{}'...", asset_id);
+
+        let txn = self.database().begin_transaction()?;
+        txn.check_asset_exists(asset_id.clone())?;
+        txn.commit()?;
+        let contract_id = ContractId::from_str(&asset_id).expect("invalid contract ID");
+        let contract = self.rgb_runtime()?.export_contract(contract_id)?;
+
+        info!(self.logger(), "Export asset contract completed");
+        Ok(contract)
+    }
+
+    /// Validate and import an RGB contract without importing asset allocations.
+    ///
+    /// This registers the contract and its public metadata locally. It does not transfer asset
+    /// units, so a newly imported contract has a zero wallet balance. Every media file declared by
+    /// the contract and not already stored by the wallet must be supplied in `media_file_paths`.
+    ///
+    /// <div class="warning">This method is meant for special usage and is normally not needed,
+    /// use it only if you know what you're doing</div>
+    #[cfg(any(feature = "electrum", feature = "esplora"))]
+    pub fn import_asset_contract(
+        &self,
+        contract: RgbContract,
+        media_file_paths: Vec<String>,
+    ) -> Result<Metadata, Error> {
+        let contract_id = contract.contract_id();
+        let asset_id = contract_id.to_string();
+        info!(self.logger(), "Importing asset contract '{}'...", asset_id);
+
+        let asset_schema: AssetSchema = contract.schema_id().try_into()?;
+        self.check_schema_support(&asset_schema)?;
+        let validation_config = ValidationConfig {
+            chain_net: self.chain_net(),
+            trusted_typesystem: asset_schema.types(),
+            ..Default::default()
+        };
+        let resolver = ContractImportResolver;
+        let valid_contract = match contract.validate(&resolver, &validation_config) {
+            Ok(contract) => contract,
+            Err(ValidationError::InvalidConsignment(e)) => {
+                error!(self.logger(), "Contract consignment is invalid: {}", e);
+                return Err(Error::InvalidConsignment);
+            }
+            Err(ValidationError::ResolverError(e)) => {
+                warn!(self.logger(), "Contract consignment could not be resolved");
+                return Err(Error::Network {
+                    details: e.to_string(),
+                });
+            }
+        };
+
+        let txn = self.database().begin_transaction()?;
+        if txn.get_asset(asset_id.clone())?.is_some() {
+            let metadata = self.get_asset_metadata_impl(&txn, asset_id)?;
+            txn.commit()?;
+            info!(self.logger(), "Import asset contract completed");
+            return Ok(metadata);
+        }
+        let mut runtime = self.rgb_runtime()?;
+
+        let declared_attachments = self.extract_attachments(&valid_contract, asset_schema);
+        let mut expected_media = BTreeMap::new();
+        for attachment in declared_attachments {
+            let digest = hex::encode(attachment.digest);
+            if let Some(expected) = expected_media.insert(digest.clone(), attachment.clone())
+                && expected.ty != attachment.ty
+            {
+                return Err(Error::InvalidAttachments {
+                    details: format!(
+                        "contract declares multiple media types for digest '{digest}'"
+                    ),
+                });
+            }
+        }
+
+        let mut provided_media = BTreeMap::new();
+        for file_path in media_file_paths {
+            let (attachment, media) = self.file_details(&file_path)?;
+            let digest = media.digest.clone();
+            let Some(expected) = expected_media.get(&digest) else {
+                return Err(Error::InvalidAttachments {
+                    details: format!("unexpected media file with digest '{digest}'"),
+                });
+            };
+            if expected.ty != attachment.ty {
+                return Err(Error::InvalidAttachments {
+                    details: format!("media type mismatch for digest '{digest}'"),
+                });
+            }
+            if provided_media
+                .insert(digest.clone(), (file_path, media))
+                .is_some()
+            {
+                return Err(Error::InvalidAttachments {
+                    details: format!("duplicate media file with digest '{digest}'"),
+                });
+            }
+        }
+
+        let mut media_to_copy = vec![];
+        for digest in expected_media.keys() {
+            let destination = self.media_dir().join(digest);
+            if destination.exists() {
+                provided_media.remove(digest);
+                continue;
+            }
+            let Some((source, media)) = provided_media.remove(digest) else {
+                return Err(Error::InvalidAttachments {
+                    details: format!("missing media file with digest '{digest}'"),
+                });
+            };
+            media_to_copy.push((source, media, destination));
+        }
+
+        let mut saved_media_paths = vec![];
+        for (source, media, destination) in media_to_copy {
+            if let Err(e) = self.copy_media_file(source, &media) {
+                if let Err(cleanup_error) = Self::cleanup_media_files(&saved_media_paths) {
+                    warn!(
+                        self.logger(),
+                        "Failed cleaning up media after contract import error: {}", cleanup_error
+                    );
+                }
+                return Err(e);
+            }
+            saved_media_paths.push(destination);
+        }
+
+        let metadata = match self.save_imported_asset_contract(
+            txn,
+            &mut runtime,
+            contract_id,
+            asset_schema,
+            valid_contract,
+            &resolver,
+        ) {
+            Ok(metadata) => metadata,
+            Err(e) => {
+                if let Err(cleanup_error) = Self::cleanup_media_files(&saved_media_paths) {
+                    warn!(
+                        self.logger(),
+                        "Failed cleaning up media after contract import error: {}", cleanup_error
+                    );
+                }
+                return Err(e);
+            }
+        };
+        info!(self.logger(), "Import asset contract completed");
+        Ok(metadata)
+    }
+
+    #[cfg(any(feature = "electrum", feature = "esplora"))]
+    fn save_imported_asset_contract(
+        &self,
+        txn: DbTxn,
+        runtime: &mut RgbRuntime,
+        contract_id: ContractId,
+        asset_schema: AssetSchema,
+        valid_contract: ValidContract,
+        resolver: &ContractImportResolver,
+    ) -> Result<Metadata, Error> {
+        runtime.import_contract(valid_contract.clone(), resolver)?;
+        let local_asset_data = self.save_new_asset_internal(
+            &txn,
+            runtime,
+            contract_id,
+            asset_schema,
+            valid_contract,
+            None,
+        )?;
+        self.update_backup_info(&txn, false)?;
+        txn.commit()?;
+
+        let initial_supply = local_asset_data.initial_supply;
+        Ok(Metadata {
+            asset_schema: local_asset_data.asset_schema,
+            initial_supply,
+            max_supply: local_asset_data.max_supply.unwrap_or(initial_supply),
+            known_circulating_supply: local_asset_data
+                .known_circulating_supply
+                .unwrap_or(initial_supply),
+            timestamp: local_asset_data.timestamp,
+            name: local_asset_data.name,
+            precision: local_asset_data.precision,
+            ticker: local_asset_data.ticker,
+            details: local_asset_data.details,
+            token: local_asset_data.token,
+            reject_list_url: local_asset_data.reject_list_url,
+        })
+    }
+
     /// Color a PSBT.
     ///
     /// <div class="warning">This method is meant for special usage and is normally not needed, use
--- a/src/wallet/rust_only.rs
+++ b/src/wallet/rust_only.rs
@@ -24,23 +24,6 @@ pub struct ColoringInfo {
     pub nonce: Option<u64>,
 }
 
-#[cfg(any(feature = "electrum", feature = "esplora"))]
-struct ContractImportResolver;
-
-#[cfg(any(feature = "electrum", feature = "esplora"))]
-impl ResolveWitness for ContractImportResolver {
-    fn resolve_witness(&self, witness_id: RgbTxid) -> Result<WitnessStatus, WitnessResolverError> {
-        Err(WitnessResolverError::ResolverIssue(
-            Some(witness_id),
-            s!("contract consignments cannot contain witness history"),
-        ))
-    }
-
-    fn check_chain_net(&self, _: ChainNet) -> Result<(), WitnessResolverError> {
-        Ok(())
-    }
-}
-
 /// Map of contract ID and list of its beneficiaries
 pub type AssetBeneficiariesMap = BTreeMap<ContractId, Vec<BuilderSeal<GraphSeal>>>;
 
@@ -143,16 +126,18 @@ impl Wallet {
             trusted_typesystem: asset_schema.types(),
             ..Default::default()
         };
-        let resolver = ContractImportResolver;
-        let valid_contract = match contract.validate(&resolver, &validation_config) {
+        let valid_contract = match contract.validate(&DumbResolver, &validation_config) {
             Ok(contract) => contract,
             Err(ValidationError::InvalidConsignment(e)) => {
                 error!(self.logger(), "Contract consignment is invalid: {}", e);
                 return Err(Error::InvalidConsignment);
             }
             Err(ValidationError::ResolverError(e)) => {
-                warn!(self.logger(), "Contract consignment could not be resolved");
-                return Err(Error::Network {
+                warn!(
+                    self.logger(),
+                    "Unexpected resolver error while validating contract consignment: {}", e
+                );
+                return Err(Error::Internal {
                     details: e.to_string(),
                 });
             }
@@ -165,35 +150,19 @@ impl Wallet {
             info!(self.logger(), "Import asset contract completed");
             return Ok(metadata);
         }
-        let mut runtime = self.rgb_runtime()?;
-
         let declared_attachments = self.extract_attachments(&valid_contract, asset_schema);
-        let mut expected_media = BTreeMap::new();
-        for attachment in declared_attachments {
-            let digest = hex::encode(attachment.digest);
-            if let Some(expected) = expected_media.insert(digest.clone(), attachment.clone())
-                && expected.ty != attachment.ty
-            {
-                return Err(Error::InvalidAttachments {
-                    details: format!(
-                        "contract declares multiple media types for digest '{digest}'"
-                    ),
-                });
-            }
-        }
+        let expected_media = declared_attachments
+            .into_iter()
+            .map(|attachment| hex::encode(attachment.digest))
+            .collect::<BTreeSet<_>>();
 
         let mut provided_media = BTreeMap::new();
         for file_path in media_file_paths {
-            let (attachment, media) = self.file_details(&file_path)?;
+            let (_, media) = self.file_details(&file_path)?;
             let digest = media.digest.clone();
-            let Some(expected) = expected_media.get(&digest) else {
+            if !expected_media.contains(&digest) {
                 return Err(Error::InvalidAttachments {
                     details: format!("unexpected media file with digest '{digest}'"),
-                });
-            };
-            if expected.ty != attachment.ty {
-                return Err(Error::InvalidAttachments {
-                    details: format!("media type mismatch for digest '{digest}'"),
                 });
             }
             if provided_media
@@ -207,7 +176,7 @@ impl Wallet {
         }
 
         let mut media_to_copy = vec![];
-        for digest in expected_media.keys() {
+        for digest in &expected_media {
             let destination = self.media_dir().join(digest);
             if destination.exists() {
                 provided_media.remove(digest);
@@ -235,14 +204,9 @@ impl Wallet {
             saved_media_paths.push(destination);
         }
 
-        let metadata = match self.save_imported_asset_contract(
-            txn,
-            &mut runtime,
-            contract_id,
-            asset_schema,
-            valid_contract,
-            &resolver,
-        ) {
+        let import_result =
+            self.save_imported_asset_contract(txn, contract_id, asset_schema, valid_contract);
+        let metadata = match import_result {
             Ok(metadata) => metadata,
             Err(e) => {
                 if let Err(cleanup_error) = Self::cleanup_media_files(&saved_media_paths) {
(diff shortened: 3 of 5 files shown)


src/utils.rs
```diff
@@ -699,9 +699,33 @@ pub struct RgbRuntime {
     stock: Stock,
     /// The wallet directory, where the lockfile for the runtime is to be held
     wallet_dir: PathBuf,
+    /// Whether dropping the runtime should persist the in-memory stock.
+    persist_on_drop: bool,
```
dcorral: Please drop `persist_on_drop`, `suppress_drop_persistence()`, `persist()` and the `Drop` guard. They are no-ops: the stock is loaded with autosave on, and `Stock::import_contract` commits its own transaction, which calls `store()` and clears the dirty flag. By the time `persist()` runs there is nothing left to persist.

src/wallet/offline.rs
```diff
@@ -1280,7 +1320,8 @@ pub trait WalletOffline: WalletBackup {
                     reserves: None,
                 };
                 if token_light.embedded_media || token_light.reserves {
-                    let runtime = self.rgb_runtime()?;
+                    let mut runtime = self.rgb_runtime()?;
+                    runtime.suppress_drop_persistence();
```
dcorral: Revert with the persistence change. Unrelated read path, and the only one of ~22 `rgb_runtime()` callers touched.

src/wallet/offline.rs
```diff
@@ -67,6 +67,46 @@ pub trait WalletOffline: WalletBackup {
         Ok(())
     }
 
+    fn extract_attachments(
```
dcorral: Is offline support required? #56 doesn't ask for it, and every other method in `rust_only.rs` is gated on `electrum`/`esplora`. Gating the new methods the same way removes this move, the `lib.rs` cfg churn and the test-const changes.

If offline stays, make this a pure move.

src/lib.rs
```diff
@@ -125,12 +126,12 @@ pub use crate::{
...
 use std::{
     cmp::{Ordering, max, min},
-    collections::{BTreeSet, hash_map::DefaultHasher},
+    collections::hash_map::DefaultHasher,
     hash::Hasher,
     num::NonZeroU32,
 };
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
```
dcorral: This and the hunks at lines 129, 172, 189, 233, 248, 260 only exist for the offline build. Not needed if the new methods are gated (see `extract_attachments` comment).

src/wallet/rust_only.rs
```diff
@@ -24,6 +24,42 @@ pub struct ColoringInfo {
     pub nonce: Option<u64>,
 }
 
+/// Result of importing an RGB asset contract without receiving allocations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ImportAssetContractResult {
```
dcorral: Drop this struct. `already_imported` only makes sense with the repair logic, which I'm asking to remove. Return `Metadata` (already computed at 255) or `()` like `save_new_asset`.

src/wallet/rust_only.rs
```diff
@@ -24,6 +24,42 @@ pub struct ColoringInfo {
...
+    /// Imported contract ID.
+    pub asset_id: String,
+    /// Whether the contract, metadata and declared media were already present before the call.
+    pub already_imported: bool,
+    /// Metadata derived from the validated contract.
+    pub metadata: Metadata,
+}
+
+struct PreparedContractMedia {
```
dcorral: One-field wrapper around a `String`. Use `BTreeMap<String, String>` or `BTreeMap<String, Vec<u8>>` if this map survives.

src/wallet/rust_only.rs
```diff
@@ -24,6 +24,42 @@ pub struct ColoringInfo {
...
+
+impl ResolveWitness for ContractImportResolver {
+    fn resolve_witness(&self, witness_id: RgbTxid) -> Result<WitnessStatus, WitnessResolverError> {
+        Err(WitnessResolverError::ResolverIssue(
+            Some(witness_id),
+            s!("contract consignments cannot contain witness history"),
+        ))
+    }
+
+    fn check_chain_net(&self, chain_net: ChainNet) -> Result<(), WitnessResolverError> {
```
dcorral: Dead code. The validator already rejects a chain-net mismatch before calling the resolver (`rgb-consensus-0.11.1-rc.11/src/validation/validator.rs:265`), and then passes `validation_config.chain_net`, which is the same value stored here. Always self-vs-self. The wrong-network test passes because of the validator, not this.

Keep only the non-panicking `resolve_witness`, make this a unit struct, return `Ok(())` here.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+    /// [`FromStr`] implementations encode and decode the ASCII-armored `RGB CONSIGNMENT` format.
+    /// It contains the public contract definition, not transfer history, wallet allocations or
+    /// proof that this wallet owns any units of the asset.
+    ///
+    /// <div class="warning">This method is meant for special usage and is normally not needed,
+    /// use it only if you know what you're doing</div>
+    pub fn export_asset_contract(&self, asset_id: String) -> Result<RgbContract, Error> {
+        info!(self.logger(), "Exporting asset contract '{}'...", asset_id);
+
+        let contract_id = ContractId::from_str(&asset_id).map_err(|error| Error::Internal {
```
dcorral: Two convention breaks:
- Parsing before the DB lookup returns `Internal` for a bad ID. Other entry points call `check_asset_exists` first (→ `AssetNotFound`) then `ContractId::from_str(..).expect("invalid contract ID")`.
- Line 141 drops the real `StockError`. `export_contract` already returns `InternalError` and `From<InternalError> for Error` exists, so just `?`.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+    /// This registers the contract and its public metadata locally. It does not transfer asset
+    /// units, and a newly imported contract therefore has a zero wallet balance.
+    ///
+    /// Contracts without media can use this method directly. If the contract declares media,
+    /// use [`import_asset_contract_with_media`](Wallet::import_asset_contract_with_media) and
+    /// provide every file that is not already present in the wallet.
+    ///
+    /// <div class="warning">This method is meant for special usage and is normally not needed,
+    /// use it only if you know what you're doing</div>
+    pub fn import_asset_contract(
```
dcorral, line 113: Do we need two methods? One import_asset_contract(contract, media_file_paths: Vec<String>) with an empty vec matches issue_asset_uda and halves the surface.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+
+        let validation_config = ValidationConfig {
+            chain_net: self.chain_net(),
+            trusted_typesystem: asset_schema.types(),
+            ..Default::default()
+        };
+        let resolver = ContractImportResolver {
+            chain_net: self.chain_net(),
+        };
+        let valid_contract = match contract.validate(&resolver, &validation_config) {
```
dcorral: Match ValidationError::InvalidConsignment / ResolverError explicitly like rust_only.rs:770 and online.rs:1185 do.

Also line 184: log the contract ID like the sibling methods ("Importing asset contract '{}'...").

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+            Err(error) => {
+                error!(self.logger(), "Contract consignment is invalid: {}", error);
+                return Err(Error::InvalidConsignment);
+            }
+        };
+
+        let declared_media = self.extract_attachments(&valid_contract, asset_schema);
+        let prepared_media = self.prepare_contract_media(&declared_media, media_file_paths)?;
+
+        let import_result: Result<bool, Error> = (|| {
```
dcorral: Please remove the closure and the contract_in_stock / asset_in_database / media_changed / already_imported logic. With the persistence flag gone there is no interruption to repair. Bare minimum is the save_new_asset shape: asset row exists → return early or error; otherwise import_contract + save_new_asset_internal + update_backup_info + commit, with ?.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+            let media_changed = self.persist_contract_media(&declared_media, &prepared_media)?;
+            let already_imported = contract_in_stock && asset_in_database && !media_changed;
+
+            if already_imported {
+                txn.commit()?;
+                return Ok(true);
+            }
+
+            if !contract_in_stock {
+                runtime.import_contract(valid_contract.clone(), &resolver)?;
```
dcorral: second valid_contract.clone() is the last use, move it.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+                    contract_id,
+                    asset_schema,
+                    valid_contract.clone(),
+                    None,
+                )?;
+            }
+
+            self.update_backup_info(&txn, false)?;
+            if !contract_in_stock {
+                runtime.persist()?;
```
dcorral: persist() is a no-op.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+            txn.commit()?;
+            Ok(false)
+        })();
+
+        // Valid media is content-addressed and deliberately survives a later stock/database
+        // failure. Keeping that durable prefix allows a retry to repair the import without the
+        // original source file.
+        let already_imported = import_result?;
+
+        let metadata = self.get_asset_metadata(asset_id.clone())?;
```
dcorral: save_new_asset_internal already returns a LocalAssetData that covers Metadata; no need to reopen DB/stock.

src/wallet/rust_only.rs
```diff
@@ -79,6 +115,369 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+        info!(self.logger(), "Import asset contract completed");
+        Ok(ImportAssetContractResult {
+            asset_id,
+            already_imported,
+            metadata,
+        })
+    }
+
+    fn prepare_contract_media(
```
dcorral: This function and everything down to line 480 re-implement media ingestion that already exists twice: file_details + copy_media_file (issuance) and the out-of-band branch of the receive path (online.rs:693-757). Both are read → hash_bytes_hex → fs::write. The crate never fsyncs outside backup, and the persist_noclobber branch at 421 is unreachable under the runtime lock.

Please follow the existing pattern and reuse cleanup_media_files

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
     assert_eq!(asset_model.schema, AssetSchema::Uda);
 }
 
+#[test]
+#[parallel]
+fn export_import_asset_contract_offline() {
```
dcorral: For all new tests: this file uses <api>_success / <api>_fail names. Suggested final set:
- export_asset_contract_success / _fail
- import_asset_contract_success (four schemas with media) / _fail (media errors, wrong network)
- a few lines in send::success for the post-transfer case

Here: move the use rgbstd::{GenesisSeal, Vout} to the top, and the ContractBuilder block (309-334) is duplicated at 684. This test disappears if the API gets gated.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+    assert_eq!(
+        recipient.wallet.export_asset_contract(asset_id).unwrap(),
+        contract
+    );
+}
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn export_import_asset_contract_success() {
```
dcorral: Same loop as import_asset_contract_with_media_success (500) minus media. Merge into one import_asset_contract_success with media. Keep the armor header asserts.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+                .unwrap(),
+            issuer.wallet.export_asset_contract(asset_id).unwrap()
+        );
+    }
+}
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn import_asset_contract_handles_post_transfer_exports() {
```
dcorral: This repeats the full send/refresh/mine cycle from send::success. Move the two useful asserts there (empty bundles/terminals on export, bundle-bearing contract → InvalidConsignment) and delete this test.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+    assert_matches!(result, Err(Error::InvalidConsignment));
+}
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn import_asset_contract_with_media_success() {
+    initialize();
+
+    let media_dir = tempfile::tempdir().unwrap();
```
dcorral: Reuse the existing fixtures (FILE_STR, tests/qrcode.png) instead of six temp files. UDA needs at most one extra file.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+    let mut issuer = get_funded_party!();
+    let recipient = get_empty_party!();
+    let asset = issuer.issue_asset_cfa(None, Some(declared_media.to_string_lossy().to_string()));
+    let contract = issuer
+        .wallet
+        .export_asset_contract(asset.asset_id.clone())
+        .unwrap();
+
+    let missing = recipient.wallet.import_asset_contract(contract.clone());
+    assert_matches!(missing, Err(Error::InvalidAttachments { .. }));
```
dcorral: 617, 626 and 637 all assert the same variant, so the test can't tell which branch fired. Check details like issue_asset_uda.rs:406.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+    let repaired = recipient
+        .wallet
+        .import_asset_contract_with_media(
+            contract,
+            vec![declared_media.to_string_lossy().to_string()],
+        )
+        .unwrap();
+    assert!(!repaired.already_imported);
+    assert_eq!(fs::read(recipient_media).unwrap(), b"declared media");
+}
```
dcorral: corrupt/repair goes with the repair logic. Rename to import_asset_contract_fail and fold the wrong-network case (667) in.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn import_asset_contract_rejects_wrong_network() {
```
dcorral: Offline scenario, but uses a funded party (docker) just for check_ticker/check_name/check_precision. Use offline_party! like line 296 and share the builder helper. Move the in-body use to the top.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+    .into_consignment();
+
+    let result = recipient.wallet.import_asset_contract(contract);
+    assert_matches!(result, Err(Error::InvalidConsignment));
+}
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn import_asset_contract_repairs_partial_persistence() {
```
dcorral: Remove with the repair logic. Also: 757 borrows the issuer's runtime to fake a state the code can't produce, and nothing asserts the stock is populated after the repair.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+        .import_asset_contract(contract)
+        .unwrap();
+    assert!(!repaired.already_imported);
+    assert_eq!(repaired.metadata.asset_schema, AssetSchema::Nia);
+}
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn import_asset_contract_recovers_after_lock_contention() {
```
dcorral: Remove. This tests the existing runtime lockfile, not this feature; every API fails the same way while held_runtime is alive.

src/wallet/test/rust_only.rs
```diff
@@ -289,6 +289,557 @@ fn save_new_asset_success() {
...
+    let repeated = competing_wallet
+        .import_asset_contract(issuer.wallet.export_asset_contract(asset.asset_id).unwrap())
+        .unwrap();
+    assert!(repeated.already_imported);
+}
+
+#[cfg(feature = "electrum")]
+#[test]
+#[parallel]
+fn export_asset_contract_requires_registered_asset() {
```
dcorral: Rename to export_asset_contract_fail.

dcorral (changes requested):
The core is right. Export uses rgb-ops' own RGB CONSIGNMENT armor, import validates network and schema, and a non-panicking resolver is an improvement for untrusted input.

The problem is scope. #56 asks for two methods; this PR is +1033 and most of it isn't the feature. Three things need to come out:

1. The persist_on_drop / persist() machinery on RgbRuntime. It's a no-op: the stock autosaves on commit inside import_contract.
2. The ~215-line media subsystem. The crate already ingests media twice with read → hash → fs::write; please follow that pattern.
3. The idempotency/repair logic and ImportAssetContractResult. With (1) gone there is nothing to repair.

Jainakin:
Thanks for the detailed review. I addressed all 25 inline threads and rebuilt the change on current `master` at `210a15f`.

The PR now contains only the two capabilities requested by #56: `export_asset_contract` and a single `import_asset_contract(contract, media_file_paths) -> Metadata` API. It uses Stock's existing autosave semantics, reuses the existing wallet media helpers, and follows the existing rust-only feature gating and error conventions.

The custom runtime persistence machinery, repair state machine, offline-build changes, result/media wrappers, duplicated media filesystem implementation, and unrelated read-path changes have been removed. The server-side diff is now 5 files and 476 additions, down from 7 files and 1,033 additions.

Validation on the current head includes all formatting checks, all four strict Clippy configurations, all four release-build configurations, 49 no-default-feature tests, 2 doc tests, 4 focused Docker-backed contract tests, and the modified end-to-end `send::success` test. The final broad run passed 312 tests with no failures after filtering only the four external multisig integration tests.

The local multisig-hub credential failure was also reproduced on untouched upstream `master`, confirming it is unrelated to this patch. Fresh GitHub Actions runs were created for `210a15f`, but GitHub currently marks them `action_required`; a maintainer needs to approve the fork workflows before the authoritative Linux matrix can run.

I have replied to each inline thread with the corresponding change. Please re-review the current head once CI is approved.

src/utils.rs
Jainakin: Removed `persist_on_drop`, `suppress_drop_persistence`, `persist`, and the custom `Drop` behavior. Contract import now relies exclusively on the existing autosaved Stock transaction.

src/wallet/offline.rs
Jainakin: Reverted this unrelated read-path change completely. `get_asset_metadata` now matches current `master`.

src/wallet/offline.rs
Jainakin: Offline build support was removed from this feature. The new resolver and APIs are gated consistently behind `electrum` or `esplora`, and the moved imports and test constants were restored.

src/lib.rs
Jainakin: Removed the offline-only cfg churn. The only remaining `lib.rs` change is the `RgbContract` re-export required by the public API.

src/wallet/rust_only.rs
Jainakin: Removed `ImportAssetContractResult`. `import_asset_contract` now returns the existing `Metadata` type; an already-known database asset returns its existing metadata.

src/wallet/rust_only.rs
Jainakin: Removed the one-field wrapper. Media matching now uses digest-keyed `BTreeMap`s with the existing `Attachment` and `Media` types.

src/wallet/rust_only.rs
Jainakin: Changed the resolver to a unit struct. `check_chain_net` now returns `Ok(())`; the validator remains responsible for network mismatch rejection, covered by the wrong-network test.

src/wallet/rust_only.rs
Jainakin: Updated both convention breaks: the database existence check now happens before invariant parsing, malformed or unknown IDs return `AssetNotFound`, and `export_contract` errors are propagated with `?`.

src/wallet/rust_only.rs
Jainakin: Collapsed the surface to one `import_asset_contract(contract, media_file_paths)` method. Callers use an empty vector when no external media is required.

src/wallet/rust_only.rs
Jainakin: Validation now explicitly matches `InvalidConsignment` and `ResolverError`, and the start log includes the derived contract ID.

src/wallet/rust_only.rs
Jainakin: Removed the repair closure and all `contract_in_stock`, `asset_in_database`, `media_changed`, and `already_imported` state logic. The path is now: existing asset returns early; otherwise Stock import, `save_new_asset_internal`, backup update, and database commit.

src/wallet/rust_only.rs
Jainakin: Removed the redundant second clone. One ownership clone remains because `RgbRuntime::import_contract` and `save_new_asset_internal` both consume the validated contract; the final value is moved into the database metadata path.

src/wallet/rust_only.rs
Jainakin: Removed the explicit `persist()` call and the entire custom persistence API.

src/wallet/rust_only.rs
Jainakin: Metadata is now constructed directly from the `LocalAssetData` returned by `save_new_asset_internal`; there is no database or Stock reopen.

src/wallet/rust_only.rs
Jainakin: Deleted the custom media persistence and fsync subsystem. Import now reuses `file_details`, `copy_media_file`, and `cleanup_media_files`, while holding the existing runtime lock across media ingestion.

src/wallet/test/rust_only.rs
Jainakin: Consolidated the new coverage into `export_asset_contract_success/fail` and `import_asset_contract_success/fail`, with the post-transfer assertions in `send::success`. `GenesisSeal` and `Vout` are imported at module scope and the shared contract builder helper is used.

src/wallet/test/rust_only.rs
Jainakin: Merged the media and no-media schema loops into one `import_asset_contract_success` test covering NIA, CFA, IFA, and UDA. The armor header and parser round-trip assertions remain in the export success test.

src/wallet/test/rust_only.rs
Jainakin: Moved the useful assertions into the existing `send::success` test: exports have empty bundles and terminals, while history-bearing input is rejected as `InvalidConsignment`. The duplicated transfer lifecycle test was removed.

src/wallet/test/rust_only.rs
Jainakin: Replaced the temporary media fixtures with the existing `FILE_STR` and `tests/qrcode.png` fixtures.

src/wallet/test/rust_only.rs
Jainakin: The failure test now checks exact, distinct details for missing, unexpected, and duplicate media, so each branch is independently verified.

src/wallet/test/rust_only.rs
Jainakin: Removed corruption and repair coverage with the repair implementation. The remaining media and wrong-network cases are consolidated under `import_asset_contract_fail`.

src/wallet/test/rust_only.rs
Jainakin: The wrong-network case now uses `offline_party!(get_test_wallet(...))` and the shared contract builder helper. The in-body RGB imports were moved to module scope.

src/wallet/test/rust_only.rs
Jainakin: Removed this repair test and the artificial borrowed-runtime state construction.

src/wallet/test/rust_only.rs
Jainakin: Removed the runtime-lock contention test; it exercised existing global lock behavior rather than this feature.

src/wallet/test/rust_only.rs
Jainakin: Renamed and consolidated this coverage under `export_asset_contract_fail`.

src/wallet/rust_only.rs
```diff
@@ -24,6 +24,23 @@ pub struct ColoringInfo {
     pub nonce: Option<u64>,
 }
 
+#[cfg(any(feature = "electrum", feature = "esplora"))]
+struct ContractImportResolver;
```
dcorral: This is unreachable. rgb-ops `Consignment::validate` rejects a contract consignment with bundles or terminals before any witness is resolved, and a contract without bundles never calls the resolver. Use the existing `DumbResolver` and drop this struct. The PR body should also stop crediting the resolver for the history rejection; rgb-ops does that.

src/wallet/rust_only.rs
```diff
@@ -79,6 +96,208 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+            ..Default::default()
+        };
+        let resolver = ContractImportResolver;
+        let valid_contract = match contract.validate(&resolver, &validation_config) {
+            Ok(contract) => contract,
+            Err(ValidationError::InvalidConsignment(e)) => {
+                error!(self.logger(), "Contract consignment is invalid: {}", e);
+                return Err(Error::InvalidConsignment);
+            }
+            Err(ValidationError::ResolverError(e)) => {
```
dcorral, line 135: Unreachable for the same reason as the resolver, and it's part of the uncovered lines codecov flags. Fine to keep for symmetry with the sibling sites, but `Error::Internal` is the honest variant here; this operation never touches the network.

src/wallet/rust_only.rs
```diff
@@ -79,6 +96,208 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+            contract_id,
+            asset_schema,
+            valid_contract,
+            None,
+        )?;
+        self.update_backup_info(&txn, false)?;
+        txn.commit()?;
+
+        let initial_supply = local_asset_data.initial_supply;
+        Ok(Metadata {
```
dcorral: This duplicates `get_asset_metadata_impl` including its `unwrap_or(initial_supply)` defaults. Call `self.get_asset_metadata_impl(&txn, asset_id)?` before `txn.commit()` instead; same transaction, no reopen, no second literal to keep in sync.

src/wallet/test/rust_only.rs
```diff
@@ -1,5 +1,44 @@
 use super::*;
 
+use rgbstd::{GenesisSeal, Vout};
+
+fn nia_contract(wallet: &Wallet, chain_net: ChainNet) -> RgbContract {
```
dcorral, line 7: Under `--no-default-features --features esplora --all-targets` this is dead code and clippy `-D warnings` fails ("function `nia_contract` is never used"). CI's esplora lint runs without `--all-targets` so it stays green there. Add `#[cfg(feature = "electrum")]` to match its only caller.

src/wallet/rust_only.rs
```diff
@@ -79,6 +96,208 @@ pub fn check_proxy_url(proxy_url: &str) -> Result<(), Error> {
...
+        if txn.get_asset(asset_id.clone())?.is_some() {
+            let metadata = self.get_asset_metadata_impl(&txn, asset_id)?;
+            txn.commit()?;
+            info!(self.logger(), "Import asset contract completed");
+            return Ok(metadata);
+        }
+        let mut runtime = self.rgb_runtime()?;
+
+        let declared_attachments = self.extract_attachments(&valid_contract, asset_schema);
+        let mut expected_media = BTreeMap::new();
```
dcorral: These two MIME checks have no counterpart in issuance or receive, and the second one compares importer-side sniffing against issuer-side sniffing, so it can only produce false rejections. Please drop both; digest matching is the contract. Same ask as my earlier comment on the media block.

dcorral (changes requested):
Four small follow-ups inline.

src/wallet/rust_only.rs
Jainakin: Addressed in 2741520. I removed ContractImportResolver and now use the existing DumbResolver. I also corrected the PR description: rgb-ops Consignment::validate rejects bundles and terminals in a contract consignment before witness resolution; the resolver is not what enforces history rejection.

src/wallet/rust_only.rs
Jainakin: Addressed in 2741520. I kept the exhaustive ResolverError arm for symmetry, but it now logs an unexpected validation resolver failure and returns Error::Internal. This import path performs no network operation.

src/wallet/rust_only.rs
Jainakin: Addressed in 2741520. The import now calls get_asset_metadata_impl on the same DbTxn before commit, removing the duplicate Metadata construction. The imported RGB runtime is explicitly dropped first because UDA metadata reopens the runtime to read token data; this avoids contending on the wallet runtime lock.

src/wallet/test/rust_only.rs
Jainakin: Addressed in 2741520. Both nia_contract and its GenesisSeal/Vout imports are now gated with cfg(feature = "electrum"). The exact Esplora all-targets check now passes: cargo clippy --workspace --all-targets --no-default-features --features esplora -- -D warnings.

src/wallet/rust_only.rs
Jainakin: Addressed in 2741520. Expected attachments are now represented only by their committed digests, and supplied files are matched by digest. Both MIME comparisons were removed; the contract-declared attachment metadata remains authoritative when the asset metadata is persisted.

Jainakin:
Addressed the latest review in 2741520 and replied inline to all five findings. The custom resolver is removed in favor of DumbResolver; the unreachable resolver failure maps to Error::Internal; metadata now comes from get_asset_metadata_impl in the same transaction after releasing the RGB runtime; the Electrum-only helper and imports are feature-gated; and media matching is digest-only with contract metadata remaining authoritative. Local validation passed formatting, all four strict Clippy feature configurations including Esplora --all-targets, 49 no-default-feature tests, 2 doc tests, 4 Docker-backed contract tests, and the modified end-to-end send::success test. No test-harness changes were committed. The four fresh workflows for 2741520 are currently action_required and need maintainer approval before the full GitHub matrix can run.

dcorral (approved):

PR #104, opened 2026-08-20
Store BDK wallet data in rgb-lib database

markuslevonyak:
Store the BDK wallet data (its `ChangeSet`) in rgb-lib's own database instead of a separate `bdk_db` file, removing the `file_store` feature.

## Motivation

Follow-up to #87: `bdk_file_store` is a development/testing database not meant for production, and its format breaks across BDK major upgrades. A rescan cannot recover from that, since revealed-but-unpaid invoice addresses have no chain history. #87 (switching to `bdk_sqlite`) was closed in favor of integrating BDK persistence into rgb-lib's existing database layer: one database technology, no coupling between sea-orm's and BDK's SQLite stacks, room for future non-SQLite backends.

## Design

- **Relational, mirroring BDK's schema one-to-one** (table names singularized, versioning via rgb-lib's migrator): BDK's own persister needs `rusqlite`, and a serialized-blob approach would be fragile across upgrades. Keeping the schema identical makes future BDK bumps a mechanical diff against upstream's `rusqlite_impl.rs`.
- **Persistence rides rgb-lib's DB transaction**: the pool has a single connection and BDK persists happen while a `DbTxn` is open (sync, broadcast, address reveal), so with single-writer SQLite a separate connection would deadlock. Bonus: BDK and RGB state now commit or roll back atomically.
- **Signing and watch-only wallets share one store**: BDK persists only the public descriptor (keys live in the in-memory signers), so both produce the identical `ChangeSet`. The separate `bdk_db_watch_only` file was a pre-1.0-BDK workaround.

## Migration

None: the old `bdk_db` file is ignored and chain data is rebuilt on the next sync. This one-time break is unavoidable; the new store is what prevents it from recurring.


src/wallet/core.rs
```diff
@@ -253,37 +250,38 @@ pub(crate) fn setup_db<P: AsRef<Path>>(wallet_dir: P) -> Result<RgbLibDatabase,
...
     desc_vanilla: String,
     watch_only: bool,
-    bdk_network: BdkNetwork,
-) -> Result<(PersistedWallet<Store<ChangeSet>>, Store<ChangeSet>), Error> {
-    let chain_net: ChainNet = wallet_data.bitcoin_network.into();
-    let mut wallet_params = BdkWallet::load()
+    bitcoin_network: BitcoinNetwork,
+) -> Result<BdkWallet, Error> {
+    let chain_net: ChainNet = bitcoin_network.into();
+    let mut load_params = BdkWallet::load()
```
zoedberg:
```suggestion
    let mut wallet_params = BdkWallet::load()
```

let's avoid unnecessary changes to keep the diff small and the blame intact

zoedberg (changes requested):
Thanks a lot for this PR, excellent work! 

This is a partial review, will report some stuff here and for some other I will just do a commit on top. Please also lint the code (job is failing)

src/wallet/core.rs
markuslevonyak: You're right. Done.

zoedberg:
Reviewed and pushed a commit on top with the changes below.

### Correctness

**1. `persist_bdk` could lose BDK changes on rollback**
It called `take_staged()` as soon as the write landed in the still-open `DbTxn`, so a rollback discarded the delta from both the DB and memory. Staged changes now move into a `WalletInternals::bdk_pending` buffer that is saved to disk (`bdk_pending.json`) and is cleared only from a `DbTxn::on_commit` callback. Re-writing is idempotent (the mapping is all upserts), so a failed write or a rollback simply gets written again next time.
The buffer is flushed to disk immediately after the broadcast applies the TX, matching where master called `bdk_wallet.persist(bdk_db)`. `setup_bdk` folds the file back in on load and removes it once its contents are durably in the DB, so a crash cannot lose the wallet's view of a TX that is already on the network.

**2. Consolidated to one persist point per transaction**
`persist_bdk` was scattered across several call sites: some public APIs persisted multiple times in one operation, and the change address revealed by `TxBuilder::finish()` was not persisted at all in the transaction that produced the PSBT. There is now a single `persist_and_commit(txn)`.

**3. `last_revealed` could go backwards**
`last_revealed` was updated with an unconditional upsert, meaning that two wallet instances on the same data directory (including a signing + watch-only pair) would drift apart. The write is now `MAX(stored, excluded)`.
Two live instances on the same data directory are still problematic though, so I added a README warning making the one-live-instance-per-data-directory constraint explicit, including for the watch-only/signer pair.

**4. No upgrade path for existing wallets**
Chain data can be rebuilt by a rescan, but revealed-address indices cannot: a rescan only restores up to the last *used* index, so an address that was only revealed would be handed out again.
`file_store` is back, behind a new `bdk_file_store_migration` feature (on by default, so the upgrade path can't be lost by omission; turn it off to drop the `bdk_file_store` dependency entirely).

**5. `u64` entity columns do not work on SQLite**
Storing them as `i64` would work until a value exceeds 2^63. Changed the affected columns to `String`, as we already do for every other `u64` field.

**6. Generated entities with sea-orm-cli**
Entities have been generated with the dedicated tool as explained in the migration README. Some fields have been manually and safely cast down (e.g. `i64` -> `u32`) to avoid continuous casts in the code, this is aligned with what we did with other pre-existing fields.

**7. Improved error handling**
`encode_tx` reported an impossible failure as `Error::IO`, errors during decode were returned as encode errors, `bdk_wallet::LoadError` is no longer reported as `Error::IO` (file-store error now impossible).

**8. Removed hand-written queries**
Replaced `Expr::cust` sites with sea-query's typed builders, so table and column names come from the generated entities and a rename becomes a compile error instead of a runtime failure. In case of the `MAX` query this would not have worked on Postgres. 

### Schema

**9. Auto-increment primary keys**
All BDK tables except `bdk_wallet` (a deliberately pinned single row) now use `pk_auto(idx)` plus a unique index on the natural key, matching the convention used by `txo`, `media` and the rest of the schema.

**10. Dropped spk cache table**
We are loading BDK with the spk cache off (also made that explicit, so a future BDK update that may change the default value doesn't go unnoticed). Therefore, since that table would always be empty, it makes sense to drop it entirely.

### Performance

**11. `bdk_tx` was written up to five times per transaction**
Separate loops upserted the same row for `txs`, `first_seen`, `last_seen` and `last_evicted`, plus a stub insert per anchor. These are now gathered into one map and written once per txid.

### Structure

**12. Flattened the load/persist mapping**
The ten single-use `load_*`/`persist_*` helpers are inlined into `load_changeset` and `persist_changeset`. Used section comments to mark the groups. The `<C: ConnectionTrait>` generic on each helper is gone as well, only `&DatabaseTransaction` was ever passed, and the module's whole point is that every write shares rgb-lib's single transaction, so the generic parameter was advertising flexibility that must not be used.

**13. Parse failures routed through InternalError::from**
Dropped the local `parse_err` helper, which deviated from the error-handling style used everywhere else in the crate.

### Tests

New: `on_commit_runs_only_when_committed`, `last_revealed_never_regresses`, `partial_tx_changeset_keeps_stored_columns`, `reveal_survives_rolled_back_txn`, `legacy_bdk_store_is_imported`, `bdk_pending_file_survives_a_crash`, `recovered_pending_is_visible_right_after_load`, `pending_tmp_file_left_by_a_crash_is_recovered`. `changeset_round_trip` now carries `u64::MAX` in `last_evicted`, so the round trip covers values above i64::MAX.
Each was verified to fail against the unfixed code, not just to pass against the fixed one.

markuslevonyak:
> Reviewed and pushed a commit on top with the changes below.

Thanks for the improvements. LGTM!

PR #114, opened 2026-09-18
fix: don't panic in send_end on extra_allocations

orbitalturtle:
When a non-dry-run send_begin spends a UTXO that also holds other assets, they other assets are moved to change as extra_allocations and persisted as non-user-driven asset transfers.
update_or_save_transfers then looks these assets up in info_contents.transfers, which only holds user-driven transfers, leading to a panic of "exists at this point".

The panic happens inside the DB transaction, so the status update is rolled back and the batch stays stuck in Initiated. Every retry of send_end on the batch, including after a restart, panics the same way.

Asset transfers that are not user-driven are extra_allocations, which by definition have no recipient and no transport endpoints to sync, so we can safely skip them as a solution.


zoedberg (changes requested):
Nice catch, thank you!

Two things:
- for bug fixes we usually prefer two commits: a first one adding only the test(s), so CI fails on it and proves the bug, and then the fix on top, so CI turning green proves the fix. It makes the PR easier to review and to bisect later
- while looking into this I think there's a slightly cleaner fix than skipping non-user-driven asset transfers. Could you amend the commit so it only contains the tests and force-push? Then I'll push my version on top and I'd like your opinion on it

orbitalturtle:
@zoedberg I agree your change is an improvement, looks good!

zoedberg (approved):

